PR #40, opened 2025-02-06
Trigger more meaningful validation error messages when trainer registration failed

sfc-gh-lmerrick:
If a user fails to register their custom `Trainer` class, or they accidentally supply the incorrect trainer type in the trainer config, validation can proceed in a difficult-to-debug manner. This PR introduces more explicit error messages that trigger in this case and can help the user understand the root cause of the issue faster.

### DIFF
--- a/arctic_training/config/trainer.py
+++ b/arctic_training/config/trainer.py
@@ -178,9 +178,7 @@ class TrainerConfig(BaseConfig):
     )
     @classmethod
     def parse_sub_config(
-        cls,
-        v: Any,
-        info: ValidationInfo,
+        cls, v: Any, info: ValidationInfo
     ) -> Union[BaseConfig, List[BaseConfig]]:
         trainer_attr_map = {
             "checkpoint": "checkpoint_engine_type",
@@ -192,7 +190,14 @@ class TrainerConfig(BaseConfig):
         }
         field_name: str = info.field_name  # type: ignore
         trainer_type: str = info.data["type"]
-        trainer_cls = get_registered_trainer(trainer_type)
+        try:
+            trainer_cls = get_registered_trainer(trainer_type)
+        except KeyError as e:
+            raise KeyError(
+                f"Unable to validate {info.field_name=} because trainer type "
+                f"'{trainer_type}' is not registered. Please register the "
+                "trainer and try again."
+            ) from e
         trainer_field_default = getattr(trainer_cls, trainer_attr_map[field_name])[0]
 
         if isinstance(v, tuple) or isinstance(v, list):
--- a/arctic_training/registry/trainer.py
+++ b/arctic_training/registry/trainer.py
@@ -54,16 +54,8 @@ def register_trainer(cls: Type["Trainer"], force: bool = False) -> Type["Trainer
     _validate_class_attribute_set(cls, "config_type")
 
     trainer_attributes = [
-        (
-            "data_factory_type",
-            get_registered_data_factory,
-            register_data_factory,
-        ),
-        (
-            "model_factory_type",
-            get_registered_model_factory,
-            register_model_factory,
-        ),
+        ("data_factory_type", get_registered_data_factory, register_data_factory),
+        ("model_factory_type", get_registered_model_factory, register_model_factory),
         (
             "checkpoint_engine_type",
             get_registered_checkpoint_engine,
@@ -130,6 +122,6 @@ def get_registered_trainer(
         trainer_name = name_or_class.name
 
     if trainer_name not in _supported_trainer_registry:
-        raise ValueError(f"{trainer_name} is not a registered Trainer.")
+        raise KeyError(f"{trainer_name} is not a registered Trainer.")
 
     return _supported_trainer_registry[trainer_name]


arctic_training/registry/trainer.py
```diff
@@ -130,6 +122,6 @@ def get_registered_trainer(
         trainer_name = name_or_class.name
 
     if trainer_name not in _supported_trainer_registry:
-        raise ValueError(f"{trainer_name} is not a registered Trainer.")
+        raise KeyError(f"{trainer_name} is not a registered Trainer.")
```
sfc-gh-caxu, line 125: It should be `ValueError`. `ValueError` suggests the function (`get_registered_trainer`) receives an invalid value. https://docs.python.org/3/library/exceptions.html#ValueError

arctic_training/config/trainer.py
```diff
@@ -192,7 +190,14 @@ def parse_sub_config(
         }
         field_name: str = info.field_name  # type: ignore
         trainer_type: str = info.data["type"]
-        trainer_cls = get_registered_trainer(trainer_type)
+        try:
+            trainer_cls = get_registered_trainer(trainer_type)
+        except KeyError as e:
+            raise KeyError(
```
sfc-gh-caxu, line 196: `ValueError`, see explanations. I think `KeyError` is rarely used and only for a local data type (e.g., a dict, set etc.)

sfc-gh-caxu (commented):
Not obvious to me whether this change is needed because the error has been raised from `get_registered_trainer`

arctic_training/registry/trainer.py
sfc-gh-lmerrick: Is the registry not a key-value mapping? If it's a key-value mapping, I believe KeyError is the more specific, and thus more useful, exception to raise.

sfc-gh-lmerrick:
> Not obvious to me whether this change is needed because the error has been raised from get_registered_trainer

In my tests, the rust-implemented [`SchemaValidator.validate_python`](https://github.com/pydantic/pydantic-core/blob/db63fecda71c6815aec7a316a17d5b8174c6659a/python/pydantic_core/_pydantic_core.pyi#L91) function driving the valiation suppressed this error and passed an `info: ValidationInfo` object to the `parse_sub_config` field validator method that was simply missing the extra fields that should have been added by the tried-but-failed previous calls to this function (i.e. no `data` object).

sfc-gh-lmerrick:
Added a test. The test fails before the changes, but succeeds after.

Previous test failure output:
```
========================================================================================= test session starts ==========================================================================================
platform linux -- Python 3.10.12, pytest-8.1.1, pluggy-1.5.0
rootdir: /scratch/tests
configfile: pytest.ini
plugins: devtools-0.12.2, anyio-4.3.0, hypothesis-5.35.1, flakefinder-1.1.0, rerunfailures-14.0, shard-0.1.2, xdist-3.6.1, xdoctest-1.0.2
collected 1 item                                                                                                                                                                                       
Running 1 items in this shard

tests/trainer/test_trainer_validation.py F                                                                                                                                                       [100%]

=============================================================================================== FAILURES ===============================================================================================
______________________________________________________________________________________ test_unregistered_trainer _______________________________________________________________________________________

tmp_path = PosixPath('/tmp/pytest-of-lmerrick/pytest-4/test_unregistered_trainer0')

    @pytest.mark.cpu
    def test_unregistered_trainer(tmp_path):
        config_dict = {
            "type": "unregistered_or_nonexistent",
            "exit_iteration": 2,
            "micro_batch_size": 1,
            "model": {
                "type": "random-weight-hf",
                "name_or_path": "HuggingFaceTB/SmolLM-135M-Instruct",
                "attn_implementation": "eager",
                "dtype": "float32",
            },
            "data": {
                "max_length": 2048,
                "sources": ["HuggingFaceH4/ultrachat_200k-truncated"],
            },
            "deepspeed": {"zero_optimization": {"stage": 0}},
            "optimizer": {"type": "cpu-adam"},
        }
        # Fails in previous implementation of `TrainerConfig.parse_sub_config`, despite
        # the implementation intending for this to succeed.
        with pytest.raises(ValueError) as ctx:
>           config = TrainerConfig(**config_dict)

tests/trainer/test_trainer_validation.py:43: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
arctic_training/config/base.py:25: in __init__
    super().__init__(**data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'arctic_training.config.trainer.TrainerConfig'>, v = 0
info = ValidationInfo(config={'title': 'TrainerConfig', 'extra_fields_behavior': 'forbid', 'validate_default': True, 'populat...1, 'gradient_accumulation_steps': 1, 'micro_batch_size': 1, 'seed': 42, 'train_iters': 0}, field_name='eval_frequency')

    @field_validator("eval_frequency", mode="after")
    def validate_eval_frequency(cls, v: int, info: ValidationInfo) -> int:
        if (
>           info.data["data"].eval_sources
            or info.data["data"].train_eval_split[1] > 0.0
        ):
E       KeyError: 'data'

arctic_training/config/trainer.py:158: KeyError
======================================================================================= short test summary info ========================================================================================
FAILED tests/trainer/test_trainer_validation.py::test_unregistered_trainer - KeyError: 'data'
========================================================================================== 1 failed in 0.72s ===========================================================================================
```

PR #41, opened 2025-02-07
Refactor Data Loading

sfc-gh-mwyatt:
Refactoring data loading so that we can support more types of input data and generally cleaning up the implementation. This PR preserves backwards compatibility.

Major changes:
- `DataSource` is overhauled and now takes a `DataSourceConfig`. However we maintain support for string input for registered data sources. An example showing the old + new way to specify data sources:
```yaml
data:
  source:
    - HuggingFaceH4/ultrachat_200k
    - type: huggingface
      dataset_name: <dataset>
      kwargs:
        revision: main
```
- `DataSource` now requires a `load` method to be defined (previously `load_fn`)
- `DataFactory` now has `load`, `tokenize`, `split_data`, and `create_dataloader` methods. This splits the data pipeline into smaller chunks to allow more opportunity for fine-grain control via callbacks
- Many `SFTDataFactory`-specific pieces of code have been moved out of the base `DataFactory` and into `SFTDataFactory`
- Many pieces of code have been moved around to make the data loading pipeline more customizable

TODO:

- [x] validate loaded data is same before/after refactor
- [ ] Add better error messages (follow up PR)
- [ ] Add unit tests (follow up PR)


arctic_training/data/factory.py
```diff
@@ -129,115 +119,95 @@ def world_size(self) -> int:
...
+            self.cache_path_args,
+            *[source.cache_path_args for source in data_source_list],
         )
+        return self.config.cache_dir / hash_str
+
+    def _get_data_sources(self, split: str) -> List["DataSource"]:
+        if split == "train":
+            data_source_configs = self.config.sources
+        elif split == "eval":
+            data_source_configs = self.config.eval_sources
```
sfc-gh-sbekman, line 149: while we have an unstable API - should we make this naming consistent? as in:
```
train_sources
eval_sources
```

just `sources` is ambiguous.

arctic_training/data/factory.py
```diff
@@ -129,115 +119,95 @@ def world_size(self) -> int:
...
+
+        data_sources = []
+        for config in data_source_configs:
+            data_source = config.data_source(data_factory=self, config=config)
+            data_sources.append(data_source)
+        return data_sources
+
+    def _truncate_data(self, dataset: DatasetType) -> DatasetType:
+        local_length = len(dataset)
+        if self.world_size != 1:
```
sfc-gh-sbekman, line 161:
```suggestion
        if self.world_size > 1:
```
less ambiguous ;)

arctic_training/data/factory.py
```diff
@@ -129,115 +119,95 @@ def world_size(self) -> int:
...
+        else:
+            raise ValueError(f"Invalid split: {split}")
+
+        data_sources = []
+        for config in data_source_configs:
+            data_source = config.data_source(data_factory=self, config=config)
+            data_sources.append(data_source)
+        return data_sources
+
+    def _truncate_data(self, dataset: DatasetType) -> DatasetType:
```
sfc-gh-lmerrick, line 159: A docstring here and/or a comment at the site in which this is called would have helped me as a reader understand faster why we're doing a truncation operation -- i.e. that we're truncating the number of examples per shard in a distributed training setting to ensure evenness across workers.

arctic_training/data/factory.py
```diff
@@ -129,115 +119,95 @@ def world_size(self) -> int:
...
+        local_length = len(dataset)
+        if self.world_size != 1:
+            data_length = torch.zeros(self.world_size).cuda()
```
sfc-gh-sbekman, line 162: should we avoid using `cuda()` as it hardcodes devices? using `to(device)` consistently?

arctic_training/data/factory.py
```diff
@@ -129,115 +119,95 @@ def world_size(self) -> int:
...
+        return dataset
 
+    @callback_wrapper("tokenize")
     @abstractmethod
-    def tokenize_fn(self, trainer: "Trainer", dataset: Dataset) -> Dataset:
-        """
-        Function to tokenize the dataset. This function should be implemented by
-        the subclass and return a Dataset of tokenized data.
-        """
+    def tokenize(
```
sfc-gh-lmerrick, line 184: For users who are working at least some of the time with pre-tokenized data, it may feel a bit more natural for the default behavior to be "tokenization off" rather than "you need to implement a tokenization function, even if it's the identity function". WDYT about having a non-abstract default implementation of simply `return dataset`?

For users working with multimodal or non-text data, the name `tokenize` may be a bit confusing, since this function is actually a general-purpose preprocessing function which can be used to apply image data augmentations and other non-tokenization steps. Maybe `process` or `transform` could serve as a more inclusive function name here.

arctic_training/data/factory.py
```diff
@@ -129,115 +119,95 @@ def world_size(self) -> int:
...
-        """
-        Returns a collate function that can be used with a Torch DataLoader.
-        """
-        raise NotImplementedError(
-            "collate_fn must be implemented by DataFactory subclass."
+    @callback_wrapper("split")
+    def split_data(
+        self, training_data: DatasetType
+    ) -> Tuple[DatasetType, Optional[DatasetType]]:
+        tmp = training_data.train_test_split(
```
sfc-gh-sbekman, line 195: `tmp` is very meh, call it `dataset` perhaps or `ds`?

arctic_training/data/factory.py
```diff
@@ -129,115 +119,95 @@ def world_size(self) -> int:
...
-        """
-        return dataset
+        return training_data, evaluation_data
+
+    @callback_wrapper("create_dataloader")
+    def create_dataloader(self, dataset: DatasetType) -> DataLoader:
+        return DataLoader(
+            dataset,
+            batch_size=self.micro_batch_size,
+            sampler=RandomSampler(dataset),
```
sfc-gh-lmerrick, line 210: If `shard == False` and `world_size > 1`, do we need a different sampler which is distributed-training-aware?

arctic_training/data/hf_source.py
sfc-gh-lmerrick, line 1: Having this submodule which can load and normalize HF datasets is really convenient really broadly -- potentially even to users who want to use a different `Trainer`. Idea for the future: add some documentation and/or helper tooling to let users leverage arctic training just for managing data sources and/or dataloading even if they're writing their own `Trainer`!

As-is, it seems a bit tricky to run the data-prep code independently, since the `DataFractory` requires that you pass a `Trainer` to initialize it.

Making dataloading and training more loosely coupled may also have other benefits for keeping codebase modular and easier to navigate and reason about as it grows.

sfc-gh-lmerrick (approved):
Thanks for this major step towards generalizing and decoupling data in arctic training for more flexibility!

sfc-gh-sbekman (commented):
part 1

sfc-gh-sbekman (commented):
part 2

arctic_training/data/factory.py
sfc-gh-mwyatt: Good point - would `DistributedSampler` be a sane default here? I'm wondering if we should make the sampler choice an option in the config somehow.

arctic_training/data/factory.py
sfc-gh-mwyatt: I removed the `@abstractmethod` but I'm having trouble identifying a method name that captures the purpose of this method. I have not worked much with multimodel training - is `process` or `transform` a pretty common name to use for this step?

arctic_training/data/hf_source.py
sfc-gh-mwyatt: Noted! I'll add this to my TODO list for future improvements

arctic_training/config/base.py
```diff
@@ -15,15 +15,21 @@
...
+from arctic_training.utils import get_local_rank
+from arctic_training.utils import get_world_size
 
 
 class BaseConfig(BaseModel):
     def __init__(self, **data):
         logger.info(f"Initializing {self.__class__.__name__}")
         super().__init__(**data)
 
+    global_rank: int = Field(default_factory=get_local_rank, exclude=True)
```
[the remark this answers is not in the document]
sfc-gh-mwyatt: You're correct. Adding this to my list of things that need to be fixed!

arctic_training/data/factory.py
```diff
@@ -65,44 +64,35 @@ def __init__(self, trainer: "Trainer", config: Optional[DataConfig] = None) -> N
...
+            if self.config.use_data_cache and cache_path.exists():
+                dataset = load_from_disk(cache_path.as_posix())
+            else:
+                dataset = self.load(data_sources, split=split)
+                dataset = self._truncate_data(dataset)
+                if self.config.use_data_cache:
+                    dataset.save_to_disk(cache_path.as_posix())
+            datasets[split] = dataset
+
+        training_data, evaluation_data = datasets["train"], datasets["eval"]
```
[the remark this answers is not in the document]
sfc-gh-mwyatt: Reason for this is because we use a for loop above to load each split:
```python
datasets = {}
for split in ("train", "eval"):
    datasets[split] = ...
```
Then directly after this unpacking line
```python
training_data, evaluation_data = self.split_data(training_data)
```

I guess we could unroll the for loop, but then it just seems like we have duplicate code for loading train and eval splits. Thoughts?

arctic_training/data/factory.py
```diff
@@ -129,115 +119,95 @@ def world_size(self) -> int:
...
-            return load_from_disk(cache_path.as_posix())
-
-        data_source_cls = get_registered_data_source(
-            name_or_class=data_source, data_factory_type=self.name
+    @property
+    def cache_path_args(self) -> Dict:
+        cache_path_args = {}
+        for field in self.config.model_fields.keys():
+            if field not in ["sources", "eval_sources"]:
+                cache_path_args[field] = getattr(self.config, field)
```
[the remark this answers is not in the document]
sfc-gh-mwyatt: Correct, the `field` value will always be a defined attribute in `self.config`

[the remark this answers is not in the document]
sfc-gh-mwyatt: Also, updated this to make it more readable

arctic_training/data/factory.py
sfc-gh-mwyatt: Yes, let's do this in a follow up PR!

tests/trainer/test_sft_trainer.py
```diff
@@ -34,7 +34,7 @@ def test_sft_trainer(tmp_path):
         },
         "data": {
             "max_length": 2048,
-            "sources": ["HuggingFaceH4/ultrachat_200k-truncated"],
+            "sources": ["HuggingFaceH4/ultrachat_200k"],
```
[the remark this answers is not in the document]
sfc-gh-mwyatt: This was an accidental inclusion. Removed!

arctic_training/data/factory.py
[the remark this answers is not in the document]
sfc-gh-sbekman: understood why they weren't used from the beginning, 

no, unrolling would definitely be bad. But what would work is an inner helper function that does all the massaging, so you end up with:

```
def massage(split):
   copy the inner loop here
   return ...
training_data = massage("train")
evaluation_data = massage("eval")
```

if you use it as a closure, no need to pass anything :) That's how I normally tackle such needs.

But if you feel that's an overkill the current proposal is fine.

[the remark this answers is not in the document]
sfc-gh-mwyatt: I think this is a good suggestion! Updated in the PR

arctic_training/data/factory.py
sfc-gh-mwyatt: Added docstring to `_truncate_data`

arctic_training/data/factory.py
sfc-gh-mwyatt: Will handle this in a follow up PR.

arctic_training/registry/data.py
```diff
@@ -88,35 +85,18 @@ def register_data_factory(
...
-
-    if data_source_name not in _supported_data_source_registry[data_factory_type]:
-        raise ValueError(
-            f"Data source {data_source_name} for data factory type"
-            f" {data_factory_type} is not registered."
-        )
 
-    return _supported_data_source_registry[data_factory_type][data_source_name]
+    if data_source_name not in _supported_data_source_registry:
+        raise ValueError(f"Data source {data_source_name} is not registered.")
```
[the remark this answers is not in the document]
sfc-gh-mwyatt: Will add in a future PR

arctic_training/config/data.py
```diff
@@ -15,49 +15,63 @@
...
+    """Base DataSource configuration."""
+
+    type: str = ""
+    """ Data source type. """
+
+    process: bool = True
+    """ Whether to process the data with the data factory `process` function (e.g., tokenization for SFTDataFactory). """
+
+    shard: bool = True
+    """ Whether to shard the data. """
```
sfc-gh-jrasley, line 49: A bit more text about what sharding means here please. I believe this always shards across DP world size? but if that's not configurable let's be explicit on what sharding means.

sfc-gh-jrasley (approved):

arctic_training/config/data.py
sfc-gh-mwyatt: Done!

sfc-gh-lmerrick:
:partying_face:

PR #42, opened 2025-02-11
fix dtype pydantic field

sfc-gh-mwyatt:
@sfc-gh-sbekman

### DIFF
--- a/arctic_training/config/model.py
+++ b/arctic_training/config/model.py
@@ -20,8 +20,8 @@ from typing import Dict
 from typing import Optional
 from typing import Type
 from typing import Union
-import torch
 
+import torch
 from pydantic import field_serializer
 from pydantic import field_validator
 
--- a/arctic_training/config/model.py
+++ b/arctic_training/config/model.py
@@ -20,6 +20,7 @@ from typing import Dict
 from typing import Optional
 from typing import Type
 from typing import Union
+import torch
 
 from pydantic import field_serializer
 from pydantic import field_validator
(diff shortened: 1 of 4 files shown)


arctic_training/config/model.py
```diff
@@ -60,8 +61,13 @@ def factory(self) -> Type["ModelFactory"]:
...
 
     @field_serializer("dtype")
-    def serialize_dtype(self, value: DType) -> str:
-        return value.value
+    def serialize_dtype(self, value: torch.dtype) -> str:
+        return str(value)
+
+    @field_validator("dtype", mode="before")
+    def validate_dtype(cls, value: Union[str, DType]) -> DType:
+        # Pydantic doesnt like the custom enum class, so we have to do this for
```
sfc-gh-sbekman, line 69:
```suggestion
        # Pydantic doesn't like the custom enum class, so we have to do this
```

sfc-gh-sbekman: FWIW, For me this has worked - a single method and not needing to change anything else - not sure if it's better:

```
    @field_validator("dtype", mode="before")
    @classmethod
    def serialize_dtype(cls, value: str) -> str:
        return DType(value).value
```

sfc-gh-sbekman (commented):
I added a possible suggestion

Weirdly the config in the test depends on `flash_attention_2` and failing

tests/config/test_model_config.py
```diff
@@ -31,6 +31,8 @@
 )
 def test_dtype_field(dtype_list):
     for dtype in dtype_list:
-        config = ModelConfig(type="sft", name_or_path="model-name", dtype=dtype)
+        config = ModelConfig(
+            type="sft", name_or_path="model-name", attn_implementation="", dtype=dtype
```
sfc-gh-sbekman, line 35: should we set the default to `""`?

sfc-gh-mwyatt: Ideally we want `flash_attention_2` to be the default, but adding that library as a dependency causes issues. Basically it expects `torch` to be installed, but there is not a good way to tell the build system to install torch first. So @sfc-gh-jrasley and I settled on this with an error message to indicate how to install flash attention: https://github.com/snowflakedb/ArcticTraining/blob/3b3d39698b0bc58f114fc8daa72a4c6c3a02ab53/arctic_training/config/model.py#L72

sfc-gh-sbekman (approved):
LGTM

I assume int8 removal was intentional.

PR #43, opened 2025-02-11
Change to sdpa as default attn_implementation

sfc-gh-mwyatt:
By default we had flash attention for `attn_implementation`. However, if the user did not have this installed it would result in an error. Adding flash attention as a dependency has its own headaches, so making the default `sdpa` to avoid seeing errors when first trying to library. @sfc-gh-sbekman


sfc-gh-sbekman (approved):
Awesome!

PR #45, opened 2025-02-11
Add Sequence Parallelism via Ulysses 

sfc-gh-sbekman:
This PR is implementing/porting Sequence Parallelism via Deepspeed Ulysses 

For PR reviewers: 

Readiness status:
- [x] Code
- [x] Docs
- [x] Tests

Related:
- WIP notes:
https://docs.google.com/document/d/1_0McUfwOVhLUJf80KM-GEuemX_l6ab2X5T0XL4LuR70/edit?tab=t.0#heading=h.bw4a5a2iioao

- Ulysses integration in Meg-DS PR:
https://github.com/snowflakedb/Megatron-DeepSpeed/commit/a2a476e11c3cee81cb551630590bf716ea2f3b8c 

Dependencies:
- [x] Making attention mechanism stackable https://github.com/huggingface/transformers/issues/36609
- [x] Allow users to pass shifted labels https://github.com/huggingface/transformers/pull/36607
- [x] Liger kernel to support shifted labels (PR merged, but waiting for a new release)


sfc-gh-sbekman:
moved StepFlopCounter to this branch if we need it later: https://github.com/snowflakedb/ArcticTraining/pull/45

arctic_training/deepspeed.py
```diff
@@ -0,0 +1,1892 @@
...
+                    **kwargs_to_pass,
+                )
+
+            torch.autograd.backward(output, incoming_grad)
+            # print(f"{i} {grad_requiring_tensor_shard.grad.numel()=}")
+            # print(f"{i} {grad_requiring_tensor_grad.numel()=}")
+            # print(f"{i} {grad_requiring_tensor_shard.grad=}")
+            # print(f"{i} {grad_requiring_tensor_grad=}")
+
+        grad_requiring_tensor_grad /= shards
```
sfc-gh-srajbhandari, line 1331: Why do we need to divide by shards?

arctic_training/trainer/sft_trainer.py
```diff
@@ -41,6 +47,148 @@ class SFTTrainer(Trainer):
...
+                        vocab_size=self.model_unwrapped.config.vocab_size
+                    )
+                    grad_requiring_tensor_key="logits"
+                    seqlen = shift_labels.shape[1]
+
+                    def loss_fn(logits=None, labels=None, shift_labels=None, vocab_size=0):
+                        if all((shift_labels == -100).squeeze()):
+                            # fake loss calculation, since CE will return nan, but grads will be set
+                            # a normal loss_fn upcasts logits to float so match it
+                            loss = (logits_shard.sum() * 0.0).float()
```
sfc-gh-srajbhandari, line 148: where is logits_shard defined?

arctic_training/trainer/sft_trainer.py
```diff
@@ -41,6 +47,148 @@ class SFTTrainer(Trainer):
...
+                        else:
+                            loss = self.model_unwrapped.loss_function(
+                                logits=logits,
+                                labels=None,
+                                vocab_size=vocab_size,
+                                shift_labels=shift_labels,
+                            )
+                        return loss
```
sfc-gh-srajbhandari, line 156: I think we should make an underlying requirement that any computation that happens inside the sequence_tiling_compute do not have dependencies across shards other than the final simple sum or average. If we make this a requirement, then we cannot take weighted sums where weights are shard dependent.

I think the easiest way to meet this requirement then is to return loss * good_items, then divide the loss by total_good_items outside the sequence_tiling_compute

arctic_training/trainer/sft_trainer.py
```diff
@@ -41,6 +47,148 @@ class SFTTrainer(Trainer):
...
+
+                    from arctic_training.deepspeed import sequence_tiling_compute
+                    loss = sequence_tiling_compute(
+                        loss_fn,
+                        seqlen,
+                        num_loss_logit_shards,
+                        kwargs_to_shard,
+                        kwargs_to_pass,
+                        grad_requiring_tensor_key, # XXX: multiple keys for the general case?
```
sfc-gh-srajbhandari, line 165: divide by `total_good_items `here and multiply by good_items in `loss_fn` defintion

arctic_training/deepspeed.py
sfc-gh-sbekman: because otherwise we were getting grad `shards` times bigger as compared to a non-sharded loss calculation. 

We discussed that a while back and you said that indeed we need to normalize by the number of shard calculations.

arctic_training/trainer/sft_trainer.py
sfc-gh-sbekman: thank you for spotting - a copy-n-paste bug - fixed now, should be:
```
loss = (logits.sum() * 0.0).float()
```

arctic_training/trainer/trainer.py
```diff
@@ -255,7 +323,9 @@ def training_horizon(self) -> int:
             raise ValueError("Train dataloader not initialized.")
         if self.config.train_iters:
             return self.config.train_iters
-        return self.config.epochs * len(self.train_dataloader) // self.config.gradient_accumulation_steps
+
+        # XXX: this was incorrect for GAS
+        return self.config.epochs * len(self.train_dataloader)  # // self.config.gradient_accumulation_steps
```
sfc-gh-sbekman, line 328: @sfc-gh-mwyatt, please confirm that my correction is kosher and I will remove comments - this PR makes partial progress on reporting and accounting with GAS>1 (enough to make `loss` and counters correct)

sfc-gh-sbekman: I wonder if it's because of SP? But then the original doesn't take into an account DP, so perhaps it needs more work?

sfc-gh-sbekman: with SP>1 `len(self.train_dataloader)` is `sp_world_size*len(original_train_dataloader)`

the training loop needs to make all the iterations -  it can't make GAS times less steps - it's only the accounting that should skip reporting any iterations that weren't at the GAS boundary.

projects/sequence-parallelism/INTEGRATION.md
```diff
@@ -0,0 +1,142 @@
...
+            # this is the original code
+            loss = self.loss(batch)
+            self.model.backward(loss)
+            ...
+
+        else:
+            # sp will do backward inside sp_fwd_bwd_loss
+            # the returned loss is already averaged across ranks
+            loss = self.sp_fwd_bwd_loss(batch)
+```
```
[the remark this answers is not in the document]
sfc-gh-sbekman: the doc is outdated - was written for the original implementation, as I mentioned in the slack post only `.py` code is ready for review.

I was planning to work on the doc, but got pulled into working on plots. Will get to it now.

arctic_training/monkey_patches.py
```diff
@@ -0,0 +1,174 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""
+A module to place monkey patches until they are upstreamed
+"""
+
+
+if 1:  # CheckpointFunctionWithCPUOffload imports
```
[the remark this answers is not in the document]
sfc-gh-sbekman: This particular monkey patch isn't meant to remain in AT. It's a hack. We need to find a proper long term solution. I was hoping pytorch core would implement it, but if not we will have to use a context manager most likely that will do this work.

projects/sequence-parallelism/INTEGRATION.md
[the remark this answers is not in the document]
sfc-gh-sbekman: Once reviewed I think we should move INTEGRATION.md to the deepspeed repo, what do you think? Since that's where the components are.

sfc-gh-sbekman:
@sfc-gh-mwyatt, docs are now ready for review.

sfc-gh-mwyatt (approved):

PR #48, opened 2025-02-13
Fix trainer double init call

sfc-gh-mwyatt:
A strange interaction was happening when calling the `TrainerConfig.trainer` property, where if the trainer class `__init__` method has an error, the `__init__` will be called twice! Removing trainer init outside of the config class fixes this problem. This property was not very useful anyway and was a vestigial piece of code.

@sfc-gh-sbekman


sfc-gh-sbekman (approved):
Amazing, thank you, Mike!

PR #54, opened 2025-02-18
Add unit tests for DataFactory

sfc-gh-mwyatt:
- Add simple unit test verifying the loaded data for single and multi-source with SFTDataFactory
- Changed `test_helpers.py` to `helpers.py` so pytest doesn't look there for tests
- Small updates to other tests, avoid writing config to file if we don't need to


tests/helpers.py
```diff
@@ -38,18 +40,33 @@ def create_model(self, model_config) -> PreTrainedModel:
...
 
+def modify_config_for_truncated_data(self):
+    self.config.kwargs["streaming"] = True  # Avoid downloading entire dataset
+    self.config.dataset_name = self.name.removesuffix(  # Set to the real dataset name
+        "-truncated"
+    )
+
+
+def sample_data_for_truncated_dataset(self, dataset: IterableDataset) -> Dataset:
+    return Dataset.from_list(list(dataset.take(20)), features=dataset.features)
```
sfc-gh-sbekman, line 51: perhaps at some point let's remove hardcoded 20, and add `items` arg with default `20`?

could those callbacks receive `kwargs` or does one have to use `partial` to configure custom params?

no rush of course, but it'd be useful down the road - I find that often I have a need be able to tweak the size of the dataset to match the test's need better.

as I mentioned elsewhere at another framework I added magic syntax sugar to the urn of the data set which allows a much easier split and select calls via just `HuggingFaceH4/ultrachat_200k-truncated:train:20`

sfc-gh-sbekman (approved):
LGTM

Left one possible suggestion - perhaps for the future.

tests/helpers.py
sfc-gh-mwyatt: We can access the config via `self.config...` in the callbacks. Let's add this when we find a need for it. I plan to continue building out the unit tests over the coming weeks 😄

PR #55, opened 2025-02-19
Improve data loading error message

sfc-gh-mwyatt:
With some data loading configurations, it's possible to filter out all of the dataset during data loading. This leads to ambiguous error messages later. Adding checks here that raise more descriptive errors when we encounter this problem.

### DIFF
--- a/arctic_training/data/sft_factory.py
+++ b/arctic_training/data/sft_factory.py
@@ -224,10 +224,17 @@ class SFTDataConfig(DataConfig):
 
 
 def filter_dataset_length(self, dataset: DatasetType) -> DatasetType:
-    return dataset.filter(
+    dataset = dataset.filter(
         lambda x: len(x["input_ids"]) <= self.config.max_length,
         num_proc=self.config.num_proc,
+        description="Filtering dataset by max length",
     )
+    if len(dataset) < 1:
+        raise ValueError(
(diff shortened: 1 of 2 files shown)


sfc-gh-sbekman (commented):
That's very useful, Mike - thank you for working on this!

Would it be ok to add a hint to the operator to why the dataset is empty when they know it's certainly isn't empty?

e.g. perhaps adding something like "Empty dataset after filtering out data that doesn't match the set requirements, e.g. `max_length` is smaller than the length of the shortest record"

Perhaps too verbose and can be phrased more succinctly but it'll help to explain like you did on slack for me.

sfc-gh-mwyatt:
> That's very useful, Mike - thank you for working on this!
> 
> Would it be ok to add a hint to the operator to why the dataset is empty when they know it's certainly isn't empty?
> 
> e.g. perhaps adding something like "Empty dataset after filtering out data that doesn't match the set requirements, e.g. `max_length` is smaller than the length of the shortest record"
> 
> Perhaps it's too verbose and can be phrased more succinctly but it'd help to explain like you did on slack for me.

Sure I think we could add these into the actual filtering step for SFTDataFactory. Let me make those changes now!

sfc-gh-sbekman (approved):
This is great now, thank you, Michael!

PR #56, opened 2025-02-20
Refactor class registry

sfc-gh-mwyatt:
While creating unit tests for the registration system, it became clear that the previous implementation was fragile in many ways and needed an overhaul. Taking this opportunity to also address feedback about requiring `@register` decorator for new classes.

Major changes:
- Replace `@register` decorator with a `RegistryMeta` metaclass for base classes
  - No longer require `@register` and classes that inherit from a ArcticTraining base class are automatically registered
  - Much more robust and concise implementation
  - `@register` is still available but is a no-op decorator and issues a deprecation warning
- Improve error messages around inability to find a registered class:

> arctic_training.exceptions.RegistryError: not-registered-type is not a registered ModelFactory. Ensure that not-registered-type is a subclass of ModelFactory and that the class has been imported before calling a get_registered_*() function. Available registered classes of type ModelFactory are: ['huggingface', 'liger']

  - Improve error messages around validation of user-defined classes:
```python
from arctic_training import ModelFactory, TokenizerConfig

class MyBadModelFactory(ModelFactory):
    config: TokenizerConfig

    def create_model(self):
        return None
```
> arctic_training.exceptions.RegistryValidationError: MyBadModelFactory must define a 'name' attribute.

> arctic_training.exceptions.RegistryValidationError: MyBadModelFactory.config must define one or more types that are a subclass of ModelConfig for the config attribute, but we found the following types are not subclasses: ['TokenizerConfig'].

> arctic_training.exceptions.RegistryValidationError: MyBadModelFactory.create_model must accept exactly {'model_config', 'self'} as parameters, but got {'self'}.

- Add unit tests for registry system
- Moves validation of inheriting classes to within each base class rather than in `arctic_training.registry.*`

Registering new classes is now automatic
old:
```python
from arctic_training import register, ModelFactory

@register
class MyModelFactory(ModelFactory):
    name = "my factory"

    def create_config(self):
        ...

    def create_model(self, model_config):
        ...
```
new:
```python
from arctic_training import ModelFactory

class MyModelFactory(ModelFactory):
    name = "my factory"

    def create_config(self):
        ...

    def create_model(self, model_config):
        ...
```


sfc-gh-sbekman (approved):
LGTM - this is massive!

In particular thank you for adding user-friendly errors, Mike.

PR #64, opened 2025-02-25
full deprecate scheduler.lr setting

sfc-gh-mwyatt:
Since launch we had `optimizer.learning_rate` and `scheduler.learning_rate`. Fully deprecating `scheduler.learning_rate` and adding an error message if a user attempts to set this config field.


sfc-gh-sbekman (approved):
Thank you for fixing, Mike.

PR #66, opened 2025-02-25
Add PEFT support

sfc-gh-mwyatt:
We had the `peft_config` input in the `ModelConfig`, but it would not properly apply the config to the model. Changes here to enable PEFT support. Also added an example and some unit tests to verify functionality of PEFT model.

### DIFF
--- a/arctic_training/config/model.py
+++ b/arctic_training/config/model.py
@@ -21,10 +21,13 @@ from typing import Optional
 from typing import Type
 from typing import Union
 
+import peft
+from peft import PeftConfig
 from pydantic import field_validator
 
 from arctic_training.config.base import BaseConfig
 from arctic_training.config.enums import DType
+from arctic_training.logging import logger
 from arctic_training.registry import get_registered_model_factory
 
 if TYPE_CHECKING:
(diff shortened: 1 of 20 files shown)


arctic_training/config/model.py
```diff
@@ -50,13 +53,32 @@ class ModelConfig(BaseConfig):
...
+    ) -> Optional[PeftConfig]:
+        if value is None:
+            return value
+
+        if "peft_type" in value:
+            peft_type = value.pop("peft_type").capitalize()
+        else:
+            logger.warning(
+                "No 'peft_type' specified in PEFT config. Defaulting to 'Lora'."
+            )
```
sfc-gh-sbekman, line 76: may I suggest that this just creates noise?

We don't want to contribute to the myriads of warnings already emitted by the sub-systems rendering the log stream useless to end user because they can't tell what's important from non-important because there are too many warnings.

If there is an issue it should assert, if there is an indication it should be info-level for those who want it.

In this particular case you're just saying that it defaults to the default, which is what most APIs have. If the default is clearly documented there is no need to tell the user about it. It's like we are not telling that the function that was called used some defaults.

Hope it makes sense.

arctic_training/config/model.py
```diff
@@ -50,13 +53,32 @@ class ModelConfig(BaseConfig):
...
+            peft_type = "Lora"
+
+        config_cls = getattr(peft, f"{peft_type}Config")
```
sfc-gh-sbekman, line 79: If a user makes a typo here or enters a non-existing name, this will fail, need to protect for this case. Normally adding a default to None and checking for it.

tests/model/test_hf_factory.py
```diff
@@ -0,0 +1,59 @@
...
+from arctic_training.config.model import ModelConfig
+
+
+@pytest.mark.parametrize(
+    "peft_type, config_cls", [("lora", LoraConfig), ("vera", VeraConfig)]
+)
+def test_peft_config(peft_type: str, config_cls: PeftConfig):
+    config_dict = {
+        "type": "random-weight-hf",
+        "name_or_path": "HuggingFaceTB/SmolLM-135M-Instruct",
```
sfc-gh-sbekman, line 33: This model is too big for when you're testing functionality and not quality. This will make CI very slow as we add more tests (manual or eventually automatic).

We can either re-use HF tiny random models or create a set of our own. Please pick one from https://huggingface.co/hf-internal-testing

You want a tiny non-random model in tests only when you test the outcome's quality and even then in most big frameworks such tests don't run on live CI because they contribute to CI taking forever and preventing fast development cycles.

sfc-gh-sbekman (changes requested):

arctic_training/config/model.py
sfc-gh-mwyatt: Removed and raise an error instead!

arctic_training/config/model.py
sfc-gh-mwyatt: Added a check against valid PEFT types

sfc-gh-sbekman (approved):
LGTM

PR #67, opened 2025-02-26
Move to smaller CI model

sfc-gh-mwyatt:
Reducing the size of the CI model to improve unit test time.


sfc-gh-sbekman (approved):
That's perfect.

Please note sometimes these models are too small for some use cases (too few layers or  positions) in those cases we can build our own - I have a ready script for doing those.

PR #68, opened 2025-02-26
info log the caching path

sfc-gh-sbekman:
when datasets/pyarrow that created the cache and loading the cache is incompatible we need to know the path to wipe out.


sfc-gh-mwyatt:
Thanks @sfc-gh-sbekman! Can we also add a similar/same log message when loading individual data sources from cache?
https://github.com/snowflakedb/ArcticTraining/blob/13f279457cc0f49b47c1ee3e4bf8dd562fc4427d/arctic_training/data/source.py#L61

sfc-gh-mwyatt (approved):

PR #70, opened 2025-02-26
Use cached environment in actions workflows

sfc-gh-mwyatt:
Setting up the python environment & dependency install for formatting and unit test workflows took a considerable amount of time (> 50% of total workflow time). Caching the environments speeds up unit test feedback for better development!

```
Formatting:   40s --> 20s
Unit-tests: 2m30s --> 58s
```

Cached environments can be seen under the repo "Actions" tab --> "Management" --> "Caches". If necessary, we can remove cached environments to force a new environment build.

For unit tests, we check the latest torch and deepspeed version and trigger a new environment build if they are updated from the cached versions.

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -24,8 +24,8 @@ jobs:
       - name: Get latest dependency versions
         id: get_latest_versions
         run: |
-          torch_version=$(python -m pip index versions torch --index-url https://download.pytorch.org/whl/cpu | awk '{print $3}' | sed 's/,//')
-          deepspeed_version=$(python -m pip index versions deepspeed | awk '{print $3}' | sed 's/,//')
+          torch_version=$(python -m pip index versions torch --index-url https://download.pytorch.org/whl/cpu | awk '{print $3}' | sed 's/[,\n]//g')
+          deepspeed_version=$(python -m pip index versions deepspeed | awk '{print $3}' | sed 's/[,\n]//g')
           versions="torch-$torch_version-deepspeed-$deepspeed_version"
           echo "versions=$versions" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


.github/workflows/unit_tests.yaml
```diff
@@ -24,8 +24,8 @@ jobs:
       - name: Get latest dependency versions
         id: get_latest_versions
         run: |
-          torch_version=$(python -m pip index versions torch --index-url https://download.pytorch.org/whl/cpu | awk '{print $3}' | sed 's/,//')
-          deepspeed_version=$(python -m pip index versions deepspeed | awk '{print $3}' | sed 's/,//')
+          torch_version=$(python -m pip index versions torch --index-url https://download.pytorch.org/whl/cpu | awk '{print $3}' | sed 's/[,\n]//g')
+          deepspeed_version=$(python -m pip index versions deepspeed | awk '{print $3}' | sed 's/[,\n]//g')
```
sfc-gh-sbekman: should we call these `latest_blah_version` to make it easier to read w/o needing to run the code? I had to run the code to understand what it was doing since `$3` is meaningless to me :)

sfc-gh-sbekman: A possible alternative that is self-documenting is:
```
python -m pip index versions torch --index-url https://download.pytorch.org/whl/cpu | perl -ne 'm|Available versions: ([\d\.]+)| && print $1'
```
but the original awk/sed is totally fine as well.

sfc-gh-sbekman (approved):
LGTM

.github/workflows/unit_tests.yaml
sfc-gh-mwyatt: Good idea! I planned to add a comment, but your perl one-liner is nicer

PR #71, opened 2025-02-27
Fix error with unit test env caching

sfc-gh-mwyatt:
We need to ensure we are still installing the latest checked out version of ArcticTraining even when using a cached environment.

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -51,5 +51,6 @@ jobs:
           pip list
       - name: Unit Tests
         run: |
+          python -m pip install ".[testing]"
           cd tests
           python -m pytest . -m "not gpu" -s -v
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -51,6 +51,5 @@ jobs:
           pip list
       - name: Unit Tests
         run: |
-          python -m pip install ".[testing]"
           cd tests
           python -m pytest . -m "not gpu" -s -v
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,11 +14,18 @@
 # limitations under the License.
 
 import os
+import sys
+from pathlib import Path
 
 import pytest
 
 
 def pytest_configure(config):
+    # allow having multiple repository checkouts and not needing to remember to rerun
+    # 'pip install .' when switching between checkouts and running tests.
+    git_repo_path = str(Path(__file__).resolve().parents[1])
+    sys.path.insert(0, git_repo_path)
+
     # TODO: Make it so that cpu and gpu tests can be run with a single command.
     # This requires some work with tearing down/setting up dist environments
     # that have not been worked out yet.


sfc-gh-sbekman:
there is a simpler solution to this (besides installing the required packages). It's better to use it instead of dev install - because it's not atomic and can be forgotten as attested by this PR. 

Instead I always use is `tests/conftest.py` that has:

```
import sys
from pathlib import Path

# allow having multiple repository checkouts and not needing to remember to rerun
# 'pip install -e .[testing]' when switching between checkouts and running tests.
git_repo_path = str(Path(__file__).resolve().parents[1])
sys.path.insert(1, git_repo_path)
```

Voila, now pytest will always use the local checkout.

sfc-gh-mwyatt:
> there is a superior solution to this (besides installing the required packages). It's better to use it instead of dev install - because the latter is not atomic and can be forgotten as attested by this PR.
> 
> Instead I always add `tests/conftest.py` that has:
> 
> ```
> import sys
> from pathlib import Path
> 
> # allow having multiple repository checkouts and not needing to remember to rerun
> # 'pip install -e .[testing]' when switching between checkouts and running tests.
> git_repo_path = str(Path(__file__).resolve().parents[1])
> sys.path.insert(1, git_repo_path)
> ```
> 
> Voila, now pytest will always use the local checkout.
> 
> and all kinds of other amazing things can be added to it - those will come later once I start working on testing.

Is there a chance we could miss changes that cause error when running `pip install`? Otherwise, I think this is a nice solution!

sfc-gh-sbekman:
> Is there a chance we could miss changes that cause error when running pip install?

I'm not sure what you mean? The proposal is for the tests to always use the local checkout - you don't ever need to do `pip install -e .` before you run tests.

If you need to install some pre-requisite packages to run tests that's a completely different issue which isn't solved by this proposal.

In CI you have to do it anyway for other required packages, so you can just add `testing` to it.

The proposed change goes beyond CI needs.

sfc-gh-sbekman:
Once you added my proposal the original proposal of this PR is no longer needed, because you have already installed all the prerequisites a few lines earlier.

sfc-gh-mwyatt:
> > Is there a chance we could miss changes that cause error when running pip install?
> 
> I'm not sure what you mean? The proposal is for the tests to always use the local checkout - you don't ever need to do `pip install -e .` before you run tests.

Sorry, I should have been more specific - There is a chance that if a PR modifies `pyproject.toml` it could break the package install (and we would want to catch that in testing). However I think we're actually safe from this as the environment cache key is determined by the contents of `pyproject.toml`:
https://github.com/snowflakedb/ArcticTraining/blob/b124945c2f0b1f5af10a7b49b50cbeb97bfd2074/.github/workflows/unit_tests.yaml#L34

sfc-gh-sbekman (approved):

sfc-gh-sbekman:
I'm not yet well familiar with the new `pyproject.toml` way of doing things, I have only done `setup.py`-based projects - I have to learn this new way, since `setup.py` is being abandoned :( and all the flexibility it provided as one can no longer do all kinds of things during project build.

PR #76, opened 2025-02-28
Fix cache path collision

sfc-gh-mwyatt:
This PR addresses two issues:
1. In some cases we were calculating the cache path incorrectly and over-writing an existing cache
2. Some of the field values used to calculate the hash path were not necessary

Additionally, cleaning up the cache path code to be easier to read and understand


sfc-gh-sbekman (approved):
thank you for improving this functionality, Mike!

PR #95, opened 2025-03-14
Multi-Replica Generation (PR 1/4)

sfc-gh-srajbhandari:
There will be 4 PRs along this line: i) Support multi-replica + TP generation within a node, ii) Support Generation Checkpointing to allow for efficient resume when replica's fail, which seems to happen once in a while, iii) Support auto-resume when a replica fails, and iv) Support multi-replica across multiple nodes.


sfc-gh-jrasley (approved):

sfc-gh-caxu (commented):
LGTM

PR #99, opened 2025-03-18
Error on yaml config duplicate keys

### LINKED ISSUE #85: multiple config entries for the same key don't error
```
data:
  max_length: 1024
  max_length: 262144
```
doesn't err and it should.

sfc-gh-mwyatt:
`pyyaml` silently overwrites values in the config for duplicate keys. Adding a custom loader that errors when duplicate keys are detected in the input config. Resolves #85

### DIFF
--- a/arctic_training/config/trainer.py
+++ b/arctic_training/config/trainer.py
@@ -43,6 +43,7 @@ from arctic_training.config.model import ModelConfig
 from arctic_training.config.optimizer import OptimizerConfig
 from arctic_training.config.scheduler import SchedulerConfig
 from arctic_training.config.tokenizer import TokenizerConfig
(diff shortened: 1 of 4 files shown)


sfc-gh-sbekman (commented):
could it report the line number like it does with other normal errors?

This would become important once we change the config to allow multiple config blocks of the same type - like 2 models / 2 optimizers

sfc-gh-sbekman (approved):
But otherwise looks great!

sfc-gh-mwyatt:
> could it report the line number like it does with other normal errors?
> 
> This would become important once we change the config to allow multiple config blocks of the same type - like 2 models / 2 optimizers. But it's also possible that some sub name could be the same in different config block types.

Added!

PR #100, opened 2025-03-18
Add help message about DeepSpeed args to ArcticTraining launcher

### LINKED ISSUE #57: Add help message about DeepSpeed launcher args to ArcticTraining CLI
When using the `arctic_training` executable we capture extra args:
https://github.com/snowflakedb/ArcticTraining/blob/5e12a1193159bbf8bd071c9365bdcf87917c7a88/arctic_training/cli.py#L26

These args are then passed to the deepspeed launcher:
https://github.com/snowflakedb/ArcticTraining/blob/5e12a1193159bbf8bd071c9365bdcf87917c7a88/arctic_training/cli.py#L38

It would be an improvement to add a help message indicating that when running `arctic_training --help`. Additionally, it might be worthwhile to see if we can validate the deepspeed args (via the deepspeed arg parser) before launching several processes that fail due to bad deepspeed args.

sfc-gh-mwyatt:
Resolves #57

### DIFF
--- a/arctic_training/cli.py
+++ b/arctic_training/cli.py
@@ -26,7 +26,7 @@ def main():
         prog="arctic_training",
         formatter_class=argparse.RawDescriptionHelpFormatter,
         epilog=textwrap.dedent(
-            """\
+            """
             DeepSpeed Args:
                 ArcticTraining uses the DeepSpeed launcher to create a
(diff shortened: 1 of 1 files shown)


sfc-gh-sbekman (approved):
Thank you for fixing, Mike

PR #101, opened 2025-03-19
Allow CLI overrides of yaml config values

sfc-gh-mwyatt:
Adding the ability to override yaml config values via the CLI:

```shell
arctic_training my_config.yaml --overrides epochs=2 model.dtype=fp32 skip_validation=True
```

Currently a bit limited in that we can only override numeric, string, or boolean type config values. We might be able to extend this to other types if we think that would be useful.

### DIFF
--- a/arctic_training/cli.py
+++ b/arctic_training/cli.py
@@ -23,6 +23,15 @@ from pathlib import Path
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument("config", type=Path, help="ArticTraining config yaml file.")
+    parser.add_argument(
+        "--overrides",
+        nargs="+",
+        help=(
+            "Override config values. Provide a list of `key=value`, where '.' is used"
+            " to indicate nested keys. For example: --overrides epochs=10"
+            " model.dtype=FP32"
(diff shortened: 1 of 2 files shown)


sfc-gh-sbekman (commented):
As a CLI API this feels a bit odd. should we perhaps just enclose all the override into quotes as a single argument? 

```
--overrides "epochs=2 model.dtype=fp32 skip_validation=True"
```
otherwise it'll quickly become a problem as one would accidentally start adding other `--` arguments in between.
```
--overrides epochs=2 model.dtype=fp32 --gpus 10 skip_validation=True
```
it's because we are used to args to be `--key value` and that `--overrides` is too far from the values. I hope it makes sense.

If you add a top level config layer then it'd become much easier to use normal argparse, as you can then do `--main.data.max_path 10 --main.optimizer.lr 0.1` etc. I'm giving `main` as an example - we can discuss what to call it. we briefly discussed that it'd be good not to have flat config files w/o an umbrella to contain them all.

I guess it could be even `--config.foo.bar value` - let's discuss tomorrow?

PR #102, opened 2025-03-19
Resolve data cache path collision bug

sfc-gh-mwyatt:
When running with multiple ranks, the local / global rank were was excluded when calculating the cache path for a the data shard on each process. This causes all ranks to write/read to the same path. Updating the configs such that `world_size`, `local_rank`, and `global_rank` are computed fields that are no longer excluded from the cache path args.


arctic_training/config/logger.py
```diff
@@ -42,22 +41,19 @@ class LoggerConfig(BaseConfig):
     @field_validator("print_output_ranks", "file_output_ranks", mode="before")
     def fill_output_ranks(cls, v):
         if v == "*":
-            return range(get_world_size())
+            return range(int(os.getenv("WORLD_SIZE", 1)))
```
sfc-gh-sbekman, line 44: why is it hardcoded when you have a helper util already?

sfc-gh-sbekman (commented):
I wonder whether instead throwing more keys into the hash, adding something like:

`global_rank-world_size-{hash}` as the path would be something more manageable if one needs to debug things? as it'd be easier to know which cache ds is which?

I don't think you need `local_rank` at all, no?

sfc-gh-mwyatt:
> I wonder whether instead throwing more keys into the hash, adding something like:
> 
> `global_rank-world_size-{hash}` as the path would be something more manageable if one needs to debug things? as it'd be easier to know which cache ds is which?
> 
> I don't think you need `local_rank` at all, no?

I agree - but we are planning to move to using DistributedSampler instead of pre-sharding the datasets. This is just to fix the bug until we make those changes (which will happen very soon)

PR #105, opened 2025-03-24
Switch to DistributedSampler

sfc-gh-mwyatt:
Previously we were pre-sharding loading datasets and saving to a cache file for each process. However, this is more cumbersome and error-prone than loading/processing data with a single process and using the DistributedSampler. Switching to use the DistributedSampler in this PR, which simplifies data loading and makes it more readable.

Other changes:
- Caching of datasets is now the default and only behavior supported
- Helper functions to detect local/shared filesystem to determine which ranks should load/process data

### DIFF
--- a/arctic_training/data/factory.py
+++ b/arctic_training/data/factory.py
@@ -94,9 +94,8 @@ class DataFactory(ABC, CallbackMixin, metaclass=RegistryMeta):
 
             dist.barrier()  # Wait for the main process to finish its preprocessing + saving to cache
 
-            self.trainer._set_seeds(
-                self.trainer.config.seed
-            )  # Reset seeds before processing data
+            # Reset seeds after may be processing data if cache didn't exist - so that main process ends up with the same RNG if the cache was there and if it wasn't, thus ensuring reproducibility.
+            self.trainer._set_seeds(self.trainer.config.seed)
             logger.info(f"Loading dataset from cache path {cache_path.as_posix()}")
             return load_from_disk(cache_path.as_posix())
 
--- a/arctic_training/data/factory.py
+++ b/arctic_training/data/factory.py
@@ -80,7 +80,7 @@ class DataFactory(ABC, CallbackMixin, metaclass=RegistryMeta):
             data_sources = self._get_data_sources(split=split)
 
             cache_path = self.cache_path(sources=data_sources, split=split)
-            if self.config.use_data_cache and cache_path.exists():
+            if cache_path.exists():
                 logger.info(f"Loading dataset from cache path {cache_path.as_posix()}")
                 return load_from_disk(cache_path.as_posix())
 
@@ -89,9 +89,8 @@ class DataFactory(ABC, CallbackMixin, metaclass=RegistryMeta):
             dataset = self.load(data_sources, split=split)
             dataset = self._truncate_data(dataset)
 
-            if self.config.use_data_cache:
-                logger.info(f"Saving dataset to cache path {cache_path.as_posix()}")
-                dataset.save_to_disk(cache_path.as_posix())
+            logger.info(f"Saving dataset to cache path {cache_path.as_posix()}")
+            dataset.save_to_disk(cache_path.as_posix())
             return dataset
 
         training_data = get_data_split("train")
--- a/arctic_training/data/factory.py
+++ b/arctic_training/data/factory.py
@@ -20,11 +20,11 @@ from typing import List
 from typing import Optional
 from typing import Tuple
 
-import torch
+import deepspeed.comm as dist
 from datasets import concatenate_datasets
 from datasets import load_from_disk
 from torch.utils.data import DataLoader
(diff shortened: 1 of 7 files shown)


tests/data/utils.py
```diff
@@ -29,13 +29,12 @@ def create_sft_data_factory(
     model_name: str,
     sources: List[str],
     eval_sources: List[str] = [],
-    cache_dir: Optional[Path] = None,
+    cache_dir: Optional[Path] = Path("/tmp/"),
```
sfc-gh-sbekman, line 32: I don't think this is a good default, `/tmp` is usually small and won't fit the cache. I think we should just require the user to set this path, and then they are aware of what they are doing.

arctic_training/data/utils.py
```diff
@@ -14,15 +14,62 @@
 # limitations under the License.
 
 import hashlib
+from functools import cache
+from pathlib import Path
 from typing import Any
+from typing import Dict
 from typing import Union
 
+import psutil
```
sfc-gh-sbekman, line 23: this needs to go into package requirements - please copy from my implementation.

sfc-gh-sbekman: https://github.com/snowflakedb/ArcticTraining/pull/45/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

arctic_training/data/factory.py
```diff
@@ -78,21 +79,22 @@ def __init__(self, trainer: "Trainer", config: Optional[DataConfig] = None) -> N
...
+
+            # If the cache path does not exist, load the data using local/global
+            # rank 0 (depending on if file system is shared across nodes).
+            if self._is_main_process_by_path(cache_path) and not cache_path.exists():
+                dataset = self.load(data_sources, split=split)
                 logger.info(f"Saving dataset to cache path {cache_path.as_posix()}")
                 dataset.save_to_disk(cache_path.as_posix())
-            return dataset
+
+            dist.barrier()  # Wait for main process to finish saving to cache
```
sfc-gh-sbekman, line 94: This is a different logic from what I implemented.

In my implementation the syncing is already done for you - The main process doing pre-processing will not return until it's done all its work and the `barrier` is already called. 

Was something invalid about my implementation? I have already prepared a whole bunch of helper utils with docs and variations that were meant to be copied into `main` as is.

arctic_training/data/factory.py
```diff
@@ -21,18 +21,19 @@
 from typing import Optional
 from typing import Tuple
 
-import torch
+import deepspeed.comm as dist
```
sfc-gh-sbekman, line 24: just an fyi comment - beware that DS's dist is a subset of torch's dist - so it's sort of confusing that sometimes we will use torch.dist and at other times' DS' dist - nothing to do here - just highlighting this

arctic_training/data/factory.py
```diff
@@ -79,21 +80,25 @@ def __init__(self, trainer: "Trainer", config: Optional[DataConfig] = None) -> N
...
+
+            # If the cache path does not exist, load the data using local/global
+            # rank 0 (depending on if file system is shared across nodes).
+            if self.is_main_process_by_path(cache_path) and not cache_path.exists():
+                dataset = self.load(data_sources, split=split)
                 logger.info(f"Saving dataset to cache path {cache_path.as_posix()}")
                 dataset.save_to_disk(cache_path.as_posix())
-            return dataset
+
+            dist.barrier()  # Wait for main process to finish saving to cache
```
sfc-gh-sbekman, line 95:
```suggestion
            dist.barrier()  # Wait for the main process to finish its preprocessing + saving to cache
```

arctic_training/data/factory.py
```diff
@@ -79,21 +80,25 @@ def __init__(self, trainer: "Trainer", config: Optional[DataConfig] = None) -> N
...
+
+            self.trainer._set_seeds(
+                self.trainer.config.seed
+            )  # Reset seeds before processing data
```
sfc-gh-sbekman, line 99:
```suggestion
            # Reset seeds before processing data
            self.trainer._set_seeds( self.trainer.config.seed)
```
no need to create new lines and reduce readability - the comment after only works functionality if autoformatter doesn't wrap it. most likely you just didn't expect it to get wrapper - let's switch to wider width soon!

sfc-gh-sbekman: also I don't think the comment is right, shouldn't it say "after processing data if at all"

but also perhaps add a longer comment to suggest why we are doing this? perhaps:
```
# Reset seeds after may be processing data if cache didn't exist - so that main process ends up with the same RNG if the cache was there and if it wasn't.
```

sfc-gh-sbekman (approved):
I suggest to merge and then I will test as it looks it'll take me quite some time to deal with all the other changes as I haven't rebased in a while.

PR #106, opened 2025-03-24
Allow local data files for huggingface data sources

sfc-gh-mwyatt:
Allows loading of locally saved HF datasets. Example:
```yaml
data:
  sources:
    - type: huggingface # explicitly specify source type
      name_or_path: /path/to/dataset-1
    - /path/to-dataset-2 # also valid
```

To support this change, also added a pydantic validator for the `sources` and `eval_sources` fields that loads the correct config type for a given data source. Previously, we just defaulted to huggingface-hosted data sources.

### DIFF
--- a/tests/data/test_hf_source.py
+++ b/tests/data/test_hf_source.py
@@ -1,0 +1,62 @@
+# Copyright 2025 Snowflake Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from pathlib import Path
+
+from datasets import Dataset
+from datasets import DatasetDict
+from datasets import load_dataset
+
+from .utils import create_sft_data_factory
+
+
+def test_huggingface_local_data_source(model_name: str, tmp_path: Path):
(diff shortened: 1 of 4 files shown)


sfc-gh-sbekman (commented):
You can make this functionality automatic with the existing API, basing on the fact that a local ds will `exist`, so the only change required is tiny:

```
if name_or_path.exists(): # local
  load_from_disk(name_or_path)
else:
  load(name_or_path)
```

much simpler and DWIM.

I have requested long time ago for `datasets` to do that automatically in `load` but it didn't happen.

sfc-gh-mwyatt:
> You can make this functionality automatic with the existing API, basing on the fact that a local ds will `exist`, so the only change required is tiny:
> 
> ```
> if name_or_path.exists(): # local
>   datasets.load_from_disk(name_or_path)
> else:
>   datasets.load(name_or_path)
> ```
> 
> much simpler and DWIM.
> 
> I have requested long time ago for `datasets` to do that automatically in `load` but it didn't happen.

Good point! I didn't do this initially because it was difficult to implement with backwards compatibility for supporting deprecated API for data sources... but as we discussed previously, we probably don't need to worry about that in these early stages of development!

sfc-gh-jrasley (approved):

tests/data/test_hf_source.py
```diff
@@ -0,0 +1,62 @@
...
+    dataset_path = tmp_path / "local_ultrachat"
+    dataset = load_dataset(
+        "HuggingFaceH4/ultrachat_200k", streaming=True, split="train_sft"
+    )
+    dataset = Dataset.from_list(list(dataset.take(20)), features=dataset.features)
+    dataset.save_to_disk(dataset_path.as_posix())
+
+    # Load saved dataset using HuggingFaceLocal data source
+    sft_data_factory = create_sft_data_factory(
+        model_name=model_name,
```
sfc-gh-sbekman, line 56: while we are changing APIs should we rename this one to `model_name_or_path` to be consistent with both:
1. HF API
2. have the mnemonic of the arg name provide self-documentation - since currently it implies that it has not to be model name only and not a local path.

PR #112, opened 2025-03-25
Add support for user-passed data splits

sfc-gh-mwyatt:
Add support for user-passed split value in data sources. This allows greater flexibility to control which split (if there are multiple splits in a dataset) and how much data is returned from each data source. For `HFDataSource` the split supports [slice splits](https://huggingface.co/docs/datasets/en/loading#slice-splits):
```yaml
data:
  sources:
    - HuggingFaceH4/ultrachat_200k:train[:20]
    - type: huggingface
      name_or_path: Open-Orca/SlimOrca
      split: test[10:50]
```

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -38,6 +38,11 @@ jobs:
         with:
           path: ~/.cache/torch_extensions
           key: torch-extensions-${{ runner.os }}-${{ steps.setup_python.outputs.python-version }}-${{ steps.get_latest_versions.outputs.versions }}
+      - name: Restore huggingface cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/huggingface
+          key: huggingface-cache-${{ runner.os }}-${{ steps.setup_python.outputs.python-version }}-${{ steps.get_latest_versions.outputs.versions }}
       - name: Setup environment
         run: |
           python -m venv .venv
--- a/arctic_training/callback/mixin.py
+++ b/arctic_training/callback/mixin.py
@@ -131,7 +131,8 @@ class CallbackMixin:
             match = callback_re.match(name)
             if match:
                 event = f"{match.group(1)}-{match.group(2).replace('_', '-')}"
-                callbacks.append((event, member))
+                if (event, member) not in callbacks:
+                    callbacks.append((event, member))
 
         return callbacks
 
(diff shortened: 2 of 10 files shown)


arctic_training/callback/mixin.py
```diff
@@ -131,7 +131,8 @@ def _get_all_callbacks(
             match = callback_re.match(name)
             if match:
                 event = f"{match.group(1)}-{match.group(2).replace('_', '-')}"
-                callbacks.append((event, member))
+                if (event, member) not in callbacks:
```
sfc-gh-mwyatt, line 134: This was a bug in the implementation that sometimes caused callbacks to be added twice.

.github/workflows/unit_tests.yaml
```diff
@@ -38,6 +38,11 @@ jobs:
         with:
           path: ~/.cache/torch_extensions
           key: torch-extensions-${{ runner.os }}-${{ steps.setup_python.outputs.python-version }}-${{ steps.get_latest_versions.outputs.versions }}
+      - name: Restore huggingface cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/huggingface
+          key: huggingface-cache-${{ runner.os }}-${{ steps.setup_python.outputs.python-version }}-${{ steps.get_latest_versions.outputs.versions }}
```
sfc-gh-mwyatt, line 45: Caching the huggingface cache to avoid repeated downloads of datasets/models used in testing

arctic_training/config/data.py
```diff
@@ -98,12 +107,19 @@ def deprecate_cache_processed_data(cls, v: Optional[bool]) -> Optional[bool]:
...
     ) -> List[DataSourceConfig]:
         """Convert string and dict input to correct subclass of DataSourceConfig. If a string is passed, "huggingface" is used as the DataSource type."""
         data_configs = []
         for config in v:
+            split = "train" if info.field_name == "sources" else "eval"
+
             # Support passing just a dataset name or path
             if isinstance(config, str):
+                # User has passed split suffix as part of the name
+                if ":" in config:
```
sfc-gh-mwyatt, line 120: This logic should probably be moved to the `HFDataSourceConfig`, but I think this is OK while we only have support for HF-style datasets

sfc-gh-sbekman (approved):
Looks awesome, Mike!

PR #113, opened 2025-03-25
Add dev flag to repeat data samples

sfc-gh-mwyatt:
In some cases for development and debugging work, we want to load a small amount of data and repeat it until we have at least 1 sample per data parallel process. Adding `expand_to_dp_size` flag to the data config to enable this behavior.

### DIFF
--- a/arctic_training/data/factory.py
+++ b/arctic_training/data/factory.py
@@ -92,6 +92,21 @@ class DataFactory(ABC, CallbackMixin, metaclass=RegistryMeta):
             # rank 0 (depending on if file system is shared across nodes).
             if self.is_main_process_by_path(cache_path) and not cache_path.exists():
                 dataset = self.load(data_sources)
+
+                if len(dataset) < self.world_size:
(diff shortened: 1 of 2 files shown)


arctic_training/data/factory.py
```diff
@@ -92,6 +92,21 @@ def load_data_sources(
             # rank 0 (depending on if file system is shared across nodes).
             if self.is_main_process_by_path(cache_path) and not cache_path.exists():
                 dataset = self.load(data_sources)
+
+                if len(dataset) < self.world_size:
```
sfc-gh-sbekman, line 96: hmm, are you sure? this is the raw data still - if we pack samples the above won't work.

sfc-gh-sbekman: I think we have to do the same once the data has been preprocessed. 

We are making an assumption that it should create at least one packed record (e.g. for 1M seqlen)

sfc-gh-sbekman (approved):

PR #116, opened 2025-03-26
Feedback and suggestions for DPO PR

sfc-gh-mwyatt:


arctic_training/data/dpo_factory.py
```diff
@@ -39,10 +40,9 @@ class DPODataConfig(DataConfig):
     max_prompt_length: int = 4096
     """ Maximum prompt length of the input sequence. """
 
-    dpo_prompt_truncation_mode: str = "keep_start"
+    dpo_prompt_truncation_mode: Literal["keep_start", "keep_end"] = "keep_start"
```
sfc-gh-mwyatt, line 43: Use `Literal` here so that pydantic will only accept those two values

arctic_training/data/dpo_factory.py
```diff
@@ -78,10 +78,11 @@ def add_bos_token_if_needed(
...
     tokens: Dict[str, List[int]],
 ) -> Dict[str, List[int]]:
+    if bos_token_id is None:
+        return tokens
+
     len_input_ids = len(tokens["prompt_input_ids"])
-    if bos_token_id is not None and (
-        len_input_ids == 0 or bos_token_id != tokens["prompt_input_ids"][0]
-    ):
+    if (len_input_ids == 0) or (bos_token_id != tokens["prompt_input_ids"][0]):
```
sfc-gh-mwyatt, line 85: Make this easier to read and match logic of `add_eos_token_if_needed` function below

arctic_training/trainer/dpo_trainer.py
```diff
@@ -146,56 +111,52 @@ def init_ref_model_config(
         return cast(ModelConfig, subconfig)
 
     @model_validator(mode="after")
-    def build_deepspeed_config(self) -> Self:
-        ds_config = self.deepspeed
-        ds_config["train_micro_batch_size_per_gpu"] = int(self.micro_batch_size * 2)
-        ds_config["train_batch_size"] = int(
+    def update_deepspeed_dpo_config(self) -> Self:
```
sfc-gh-mwyatt, line 114: No need to copy all of the `build_deepspeed_config` from the base `TrainerConfig` class, just modify the values we want to modify

arctic_training/trainer/dpo_trainer.py
```diff
@@ -146,56 +111,52 @@ def init_ref_model_config(
...
-                "stage": 2,
-                "stage3_param_persistence_threshold": 1e4,
-                "stage3_max_live_parameters": 3e7,
-                "stage3_prefetch_bucket_size": 3e7,
-                "memory_efficient_linear": False,
-            },
+        return self
+
+    @model_validator(mode="after")
+    def build_ref_model_deepspeed_config(self) -> Self:
```
sfc-gh-mwyatt, line 128: Bring code from `get_eval_ds_config` function into a validator so that the config is setup before runtime. This will help any future debugging as the config when we initialize the `DPOTrainer` will be the config that it runs with (rather than modifying things during model setup)!

arctic_training/trainer/dpo_trainer.py
```diff
@@ -214,22 +180,19 @@ class DPOTrainer(Trainer):
...
     optimizer_factory: FusedAdamOptimizerFactory
     scheduler_factory: HFSchedulerFactory
     tokenizer_factory: HFTokenizerFactory
+    ref_model: torch.nn.Module
+    liger_dpo_loss: Optional[LigerFusedLinearDPOLoss] = None
+
     callbacks = [
         ("post-init", init_ref_model),
         ("post-init", init_liger_dpo_loss),
-        post_loss_log_cb,
```
sfc-gh-mwyatt, line 226: These are part of the base `Trainer` class and do not need to be added again

arctic_training/trainer/sft_trainer.py
```diff
@@ -28,13 +27,7 @@
...
 from arctic_training.tokenizer.hf_factory import HFTokenizerFactory
 from arctic_training.trainer.trainer import Trainer
-
-
-def to_device(batch: Dict, device: str) -> Dict:
-    output = {}
-    for k, v in batch.items():
-        output[k] = v.to(device)
-    return output
+from arctic_training.trainer.utils import to_device
```
sfc-gh-mwyatt, line 30: Moved this to a new `utils` module because we use it for mulitple trainers

examples/dpo.yaml
sfc-gh-mwyatt, line 1: Removed this since we have examples in `projects/`

sfc-gh-jrasley (approved):

PR #117, opened 2025-03-26
Update bird_config.yaml

sfc-gh-jrasley:


sfc-gh-mwyatt (approved):
Is `dev_20240627` significant? Seems like an oddly specific name

sfc-gh-jrasley:
> Is `dev_20240627` significant? Seems like an oddly specific name

it's the inner path of the data, i just downloaded it and saw it as well

sfc-gh-mwyatt:
> 
> it's the inner path of the dev dataset, i just downloaded it and saw it as well

Ahh ok that makes sense! Thanks

PR #121, opened 2025-03-27
Add `cache_fs_type` field

### LINKED ISSUE #87: [feature request] add `data.cache_fs_type` config entry
I added `cache_fs_type: auto` logic to automatically figure out whether the cache is on a local or shared fs and build the cache correspondingly, but I'm concerned there will be cases where it might fail and so I think we should give the user a way to override and set the type of cache fs explicitly, so I propose this new config entry:

```
  cache_fs_type: [auto|local|shared]
```
which would map like:
```
data:
  use_data_cache: true
  cache_processed_data: true
  cache_dir: /data-fast/data-cache/stas
  cache_fs_type: [auto|local|shared]
```
with `auto` being the default.

sfc-gh-mwyatt:
Users can now override the cache filesystem type if ArcticTraining is not detecting it correctly. The file system type determines which ranks will download and process the data (i.e., local: all local rank 0; shared: only global rank 0). If we detect this incorrectly, then some processes will have no data and training will break.

Defaults to `auto` and keeps current behavior unless user specifies `local` or `shared`

Resolves #87

### DIFF
--- a/arctic_training/config/data.py
+++ b/arctic_training/config/data.py
@@ -17,6 +17,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Dict
 from typing import List
+from typing import Literal
 from typing import Optional
 from typing import Tuple
 from typing import Type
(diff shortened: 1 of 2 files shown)


sfc-gh-sbekman (approved):
Super!

PR #125, opened 2025-03-31
Training metrics output to W&B

sfc-gh-mwyatt:
Modified `Metrics` to collect summary values in a dictionary that can be accessed by `self.metrics.summary_dict`. This dict is then reported to W&B.

Additionally:
- Moved W&B integration out of callbacks to align with feedback about not having callbacks as part of base class definitions
- Sets W&B output directory to the logging output directory (unless the user has explicitly set the `WANDB_DATA_DIR` environment variable)


arctic_training/trainer/trainer.py
```diff
@@ -313,6 +317,8 @@ def epoch(self) -> None:
 
             if self.train_batch_idx % self.config.train_log_iter_interval == 0:
                 self.metrics.print_summary()
+                if self.wandb_experiment is not None:
+                    self.wandb_experiment.log(self.metrics.summary_dict)
```
sfc-gh-sbekman, line 321: we want to tell wandb which step we are logging explicitly:
```suggestion
                    self.wandb_experiment.log(self.metrics.summary_dict, step=self.model.global_steps)
```

sfc-gh-sbekman: this allows us to e.g. skip iter 1 as it massively impacts the plot on many metrics and makes the plot more difficult to read.

that way we could go back to logging stats for iter 1 for the normal logging, but skip it for wandb.

sfc-gh-sbekman (approved):

PR #127, opened 2025-04-01
Small improvements / fixes for metrics logging

sfc-gh-mwyatt:
- Move output log files to a `logs/` subdir
- Ensure only global rank 0 is logging to W&B / printing
- Remove `step` and `epoch` from W&B reported values to avoid plots like step vs step
- Move `epochs` to end of logging line
- Reduce precision to more reasonable values for reported times
- Add human-readable formatting util functions

### DIFF
--- a/arctic_training/config/logger.py
+++ b/arctic_training/config/logger.py
@@ -53,7 +53,7 @@ class LoggerConfig(BaseConfig):
 
     @property
     def log_file(self) -> Path:
-        return self.output_dir / f"rank_{self.local_rank}.log"
+        return self.output_dir / "logs" / f"rank_{self.local_rank}.log"
 
     @property
     def file_enabled(self) -> bool:
--- a/arctic_training/trainer/trainer.py
+++ b/arctic_training/trainer/trainer.py
@@ -322,7 +322,7 @@ class Trainer(ABC, CallbackMixin, metaclass=RegistryMeta):
                 self.metrics.print_summary()
                 if (
                     self.global_rank == 0
-                    and self.train_batch_idx > 0
+                    and self.train_batch_idx > 1  # first iter is a massive outlier
                     and self.wandb_experiment is not None
                 ):
                     self.wandb_experiment.log(
--- a/arctic_training/trainer/trainer.py
+++ b/arctic_training/trainer/trainer.py
@@ -315,11 +315,7 @@ class Trainer(ABC, CallbackMixin, metaclass=RegistryMeta):
                     and self.wandb_experiment is not None
                 ):
                     self.wandb_experiment.log(
-                        {
-                            k: v
-                            for k, v in self.metrics.summary_dict.items()
-                            if k != "iter"
-                        },
+                        {k: v for k, v in self.metrics.summary_dict.items() if k != "iter"},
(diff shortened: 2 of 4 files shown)


arctic_training/config/logger.py
```diff
@@ -53,7 +53,7 @@ def set_wandb_output_dir(self) -> Self:
 
     @property
     def log_file(self) -> Path:
-        return self.output_dir / f"rank_{self.local_rank}.log"
+        return self.output_dir / "logs" / f"rank_{self.local_rank}.log"
```
sfc-gh-sbekman, line 56: is there a need to `mkdir logs` somewhere? or will this sort it out dynamically?

arctic_training/trainer/trainer.py
```diff
@@ -315,11 +315,15 @@ def epoch(self) -> None:
...
             self.metrics.restart_timer("iter")
 
-            if self.train_batch_idx % self.config.train_log_iter_interval == 0:
+            if (
+                self.config.train_log_iter_interval != 0
+                and self.train_batch_idx % self.config.train_log_iter_interval == 0
+            ):
                 self.metrics.print_summary()
-                if self.wandb_experiment is not None:
+                if self.global_rank == 0 and self.wandb_experiment is not None:
```
sfc-gh-sbekman, line 323: and we probably want to skip iter=1 for wandb, since it'll skew many plots with many first metrics being huge outliers - the first iteration is much slower than the rest.

we want it for normal print.

arctic_training/trainer/trainer.py
```diff
@@ -315,11 +315,15 @@ def epoch(self) -> None:
...
                     self.wandb_experiment.log(
-                        self.metrics.summary_dict, step=self.model.global_steps
+                        self.metrics.get_summary_dict(exclude=["iter", "epoch"]),
```
sfc-gh-sbekman, line 325: this is sort problematic naming-wise - the user doesn't know from this API if they get a copy or the actual dict and whether it's safe to modify it. perhaps `summary_dict_copy`? (not sure `get_` is helpful here, since there is no `set_`)

arctic_training/config/logger.py
sfc-gh-mwyatt: Yes, the directory will be created by the logger

arctic_training/trainer/trainer.py
```diff
@@ -320,9 +320,17 @@ def epoch(self) -> None:
                 and self.train_batch_idx % self.config.train_log_iter_interval == 0
             ):
                 self.metrics.print_summary()
-                if self.global_rank == 0 and self.wandb_experiment is not None:
+                if (
+                    self.global_rank == 0
+                    and self.train_batch_idx > 0
```
sfc-gh-sbekman, line 325:
```suggestion
                    and self.train_batch_idx > 1  # first iter is a massive outlier
```

sfc-gh-sbekman: iteration count starts from 1 and added a note for the future to why we are excluding iter 1.

sfc-gh-sbekman (approved):
Thank you, Mike

PR #138, opened 2025-04-09
Add Arctic Embed minimal example

sfc-gh-lmerrick:


projects/arctic_embed/src/arctic_embed/core/biencoder_model.py
```diff
@@ -26,19 +26,18 @@
...
-from torch import nn
 from transformers import PreTrainedModel
 
 PoolingOption = Literal["first_token", "last_token", "mean"]
 
 logger = logging.getLogger(__name__)
 
 
-class Biencoder(nn.Module):
+class Biencoder(PreTrainedModel):
```
sfc-gh-lmerrick, line 36: This is another tying fix that also now has importance due to the Metrics system expecting a PreTrainedModel where promised

arctic_training/trainer/trainer.py
```diff
@@ -297,7 +300,8 @@ def epoch(self) -> None:
         self.metrics.start_timer("iter")
         for batch in self.train_batches:
             self.train_batch_idx += 1
-            self.metrics.record("seqlen", len(batch["input_ids"][0]))
+            with suppress(Exception):  # Batch may not be dict with `input_ids` key.
```
sfc-gh-mwyatt, line 303: Will this suppress all exceptions? I could imagine if there is an error within `metrics.record` function this could silently hide that. But we have plans to overhaul the metrics class, so maybe this OK for now!

sfc-gh-lmerrick: Yeah we could catch `KeyError`, `TypeError`, and `IndexError` and probably catch most of the issues this line will have with trying to index by a string value into an object that could be any type and have any set of keys, but it feels like we can't easily anticipate all of the errors that might be raised by trying `batch["input_ids"][0]` on a `batch` object that is allowed to be of any user-defined type.

sfc-gh-lmerrick: Oh, we can split this into multiple lines and suppress errors for only certain lines, though.

sfc-gh-lmerrick: More verbose, but wdyt about this?

```python
seqlen = None
with suppress(Exception):
    seqlen = len(batch["input_ids"][0])
if seqlen is not None:
    len(batch["input_ids"][0]), seqlen)
```

PR #144, opened 2025-04-10
Automatically select CPU Adam when offloading optimizer

sfc-gh-mwyatt:


arctic_training/config/trainer.py
```diff
@@ -334,6 +334,12 @@ def validate_single_checkpoint_resume(self) -> Self:
         assert sum(resume_checkpoint_values) <= 1, "Only one checkpoint can auto resume."
         return self
 
+    @model_validator(mode="after")
+    def set_optimizer_for_cpu_offload(self) -> Self:
+        if self.deepspeed.get("cpu_offload", {}).get("device", None) == "cpu" and self.optimizer.type == "fusedadam":
+            self.optimizer.type = "cpu_adam"
```
sfc-gh-sbekman, line 340: should it assert if a user has an explicit `self.optimizer.type` set to the wrong value? like:

```
zero_optimization .offload_optimizer.cpu_offload is configured therefore optimizer.type=cpu_adam is required, but the config file has optimizer.type=... . Please either set it to `cpu_adam` or remove the explicit config and it'll be set automatically`
```

PR #152, opened 2025-04-11
an optional mem profiler

sfc-gh-sbekman:
@sfc-gh-mwyatt, let's make this configurable?

3 settings:
1. at which level to bracket the profiling - at the moment I added "e2e" (all of it) vs "step" - only around the step
2. how many items to store - currently between 100k and 1m (hardcoded at the moment) `torch.cuda.memory._record_memory_history(max_entries=100_000)`
3. where to save the results 

Actually probably should `s/step/epoch/` as it's measuring at the level of epoch.

we can discuss the naming next and/or feel free to push directly into this PR.

### DIFF
--- a/arctic_training/config/trainer.py
+++ b/arctic_training/config/trainer.py
@@ -139,6 +139,9 @@ class TrainerConfig(BaseConfig):
     mem_profiler_dir: Path = Field(default_factory=lambda data: data["logger"].output_dir / "mem-prof")
     """ Path to save memory profiling results. Defaults to `logger.output_dir/mem-prof`. """
 
+    mem_profiler_max_entries: int = Field(default=100_000, ge=1)
+    """ Maximum number of entries to store in the memory profiler. """
+
     @model_validator(mode="after")
     def init_dist(self) -> Self:
         get_accelerator().set_device(self.local_rank)
--- a/arctic_training/config/trainer.py
+++ b/arctic_training/config/trainer.py
@@ -133,6 +133,12 @@ class TrainerConfig(BaseConfig):
     overfit_first_batch: bool = False
     """ Train only on repetitions of the first training batch. Useful for development. """
 
+    mem_profiler: Literal[None, "step", "e2e"] = None
+    """ Enable memory profiling. """
+
+    mem_profiler_dir: Path = Field(default_factory=lambda data: data["logger"].output_dir / "mem-prof")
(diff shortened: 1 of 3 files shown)


sfc-gh-mwyatt:
@sfc-gh-sbekman added config fields to control this feature. Do you feel these should go under a `metrics` sub-config when we make the training metrics configurable as well?

Also, it seems there are several directories where we output values (i.e., data cache, checkpoint, logs, W&B, mem profiler). I think it might be worth following up this PR to add a `base_output_dir` (we can work on the naming) that can be used to control where all outputs go, and then we can override if a user sets a specific path (e.g., `config.wandb.output_dir`)

sfc-gh-sbekman:
> @sfc-gh-sbekman added config fields to control this feature. Do you feel these should go under a `metrics` sub-config when we make the training metrics configurable as well?

No, this would belong into `debug` - perhaps we can allocate this top-level namespace?

> Also, it seems there are several directories where we output values (i.e., data cache, checkpoint, logs, W&B, mem profiler). I think it might be worth following up this PR to add a `base_output_dir` (we can work on the naming) that can be used to control where all outputs go. For example, if a user does not define a `config.wandb.output_dir` then it will default to `config.base_output_dir / "wandb"`.

Hmm, I'm having a dejavu - haven't we discussed this already - but regardless `yes` to your proposal.

sfc-gh-mwyatt:
> I think this would belong into debug - perhaps we can allocate this top-level namespace?

This sounds good. I think this should be a follow-up PR. I'll add it as an issue

> it also has a 3rd config value - how many items to store - currently between 100k and 1m (hardcoded)

Added in the latest commit

> Hmm, I'm having a dejavu - haven't we discussed this already - and I thought it was already implemented - but regardless, it's a yes to your proposal.

We have this in that we use the `logger.output_dir` as the base dir, but we should add a `base_output_dir` at the top level that controls all output directories. For example, if we set `base_output_dir` and no other paths, we would expect to find `logs`, `wandb`, `checkpoint`, etc. sub-directories at that path!

sfc-gh-mwyatt (approved):

sfc-gh-sbekman:
> We have this in that we use the logger.output_dir as the base dir, but we should add a base_output_dir at the top level that controls all output directories. For example, if we set base_output_dir and no other paths, we would expect to find logs, wandb, checkpoint, etc. sub-directories at that path!

+1

PR #153, opened 2025-04-11
train iter log: add mem metrics

sfc-gh-sbekman:
@sfc-gh-mwyatt, should we make this configurable?

Perhaps:
```
train_log_metrics_mem_stats: true
```
This version isn't costly performance wise, so I didn't add "debug_" to the naming - should be perfectly fine for production use - so in retrospect, perhaps we should just always log it and not have it configurable?

### DIFF
--- a/arctic_training/debug.py
+++ b/arctic_training/debug.py
@@ -1,0 +1,54 @@
+# Copyright 2025 Snowflake Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


arctic_training/debug.py
```diff
@@ -0,0 +1,67 @@
...
+
+import torch.distributed as dist
+from deepspeed.accelerator import get_accelerator
+
+can_run_pynvml = True
+try:
+    import pynvml
+
+    pynvml.nvmlInit()
+except Exception:
```
sfc-gh-mwyatt, line 27:
```suggestion
except (ImportError, ModuleNotFoundError):
```

sfc-gh-mwyatt (approved):

PR #156, opened 2025-04-14
Fix for license check hook adding multiple license

sfc-gh-mwyatt:
By default pre-commit runs the pre-commit hooks on multiple parallel processes (https://pre-commit.com/#hooks-require_serial). Setting `require_serial` to true for our license check hook to prevent the adding of the license text multiple times to newly added files.


sfc-gh-sbekman (approved):
super!

Issue #158, opened 2025-04-14
better yaml format / indentation reporting


sfc-gh-sbekman:
The pure yaml formatting mistakes aren't easy to trace to where they happen.
e.g. I mistakenly ended up with:
```
model:
type: "liger"
```
because that's how vscode uncomments things mismatching manual commenting out - I didn't ask it to remove whitespace.
but that's not the point, the point is that the error doesn't tell the user where the problem is:
```
  File "/usr/local/lib/python3.10/dist-packages/yaml/parser.py", line 438, in parse_block_mapping_key
    raise ParserError("while parsing a block mapping", self.marks[-1],
yaml.parser.ParserError: while parsing a block mapping
  in "run-dp1-sp8.yml", line 1, column 1
expected <block end>, but found '<block mapping start>'
  in "run-dp1-sp8.yml", line 34, column 3
```
at the very least it should report the correct line which was 21 in this example

additionally, I feel these formatting / indentation errors are caught too late after waiting for deepspeed and processes to launch.
I was wondering if we first could quickly run a standalone yaml linter immediately in the launcher's argparse and fail if it fails? it should then give the correct line number and save the operator wasted time waiting for a bunch of other things to launch. 

and that would kill 2 birds in one stone.

PR #159, opened 2025-04-14
Early catch config errors

sfc-gh-mwyatt:
Addresses #158 

If we make a mistake in our yaml config like below, we now catch this error before launching all processes. This gives faster feedback on config problems:

Bad config yaml:
<pre>
type: sft
micro_batch_size: 36

model:
<b>type: "liger" # Accidental dedent</b>
  name_or_path: Felladrin/Llama-160M-Chat-v1
  attn_implementation: sdpa

data:
  sources:
    - HuggingFaceH4/ultrachat_200k:train
  max_length: 1024
</pre>

Error:
```
yaml.parser.ParserError: while parsing a block mapping
  in "repro.yaml", line 1, column 1
expected <block end>, but found '<block mapping start>'
  in "repro.yaml", line 6, column 3
```

### DIFF
--- a/arctic_training/cli.py
+++ b/arctic_training/cli.py
@@ -22,6 +22,8 @@ from pathlib import Path
 
 
 def main():
+    from arctic_training.config.trainer import get_config
+
     parser = argparse.ArgumentParser(
         prog="arctic_training",
         formatter_class=argparse.RawDescriptionHelpFormatter,
@@ -44,6 +46,9 @@ def main():
 
     if not args.config.exists():
         raise FileNotFoundError(f"Config file {args.config} not found.")
+
+    # Early catch any config errors
+    _ = get_config(args.config)
 
     exe_path = shutil.which("arctic_training_run")
 


sfc-gh-sbekman:
the line number doesn't match still - it should say line 5 to match your example

perhaps there is a different linter that reports things better?

sfc-gh-mwyatt:
> the line number doesn't match still - it should say line 5 to match your example
> 
> perhaps there is a different linter that reports things better?
> 
> and/or it'd have been useful to also dump the offensive line?

That's because the following is a valid yaml, where `model` gets the value `None`:
```yaml
type: sft
micro_batch_size: 36

model:
type: "liger"
```

But we can extend out custom yaml loader to report raise errors on empty values:
```python
import yaml

class UniqueKeyLoader(yaml.SafeLoader):
    def construct_mapping(self, node, deep=False):
        mapping = set()
        for key_node, _ in node.value:
            key = self.construct_object(key_node, deep=deep)
            if key in mapping:
                raise ValueError(f"Duplicate '{key}' key found in YAML on line {key_node.start_mark.line + 1}.")
            mapping.add(key)
        return super().construct_mapping(node, deep)

    def process_empty_scalar(self, mark):
        raise ValueError(f"Empty value found in YAML on line {mark.line + 1}.")

example = """\
type: sft
micro_batch_size: 36

model:
type: "liger"
"""

data = yaml.safe_load(example)
print("Successfully loaded yaml:", data)

data = yaml.load(example, Loader=UniqueKeyLoader)
```

This outputs:
```
❯ python yaml_fail.py
Successfully loaded yaml: {'type': 'liger', 'micro_batch_size': 36, 'model': None}
Traceback (most recent call last):
  File "/code/users/mwyatt/llama-4-eval/yaml_fail.py", line 27, in <module>
    data = yaml.load(example, Loader=UniqueKeyLoader)
  ...
  File "/code/users/mwyatt/llama-4-eval/yaml_fail.py", line 14, in process_empty_scalar
    raise ValueError(f"Empty value found in YAML on line {mark.line + 1}.")
ValueError: Empty value found in YAML on line 4.
```

Updated the PR to include this change.

sfc-gh-sbekman:
Yeah, as you said this is not helping and makes things even more confusing, since there is no problem with `model: ` but with subsequent line.

I didn't think that the misformatted yaml example we work with is a valid yaml when there is no schema. Thank you for pointing that out.

OK, how about we go back then to the normal parser stage and figure out how to get it to report the correct line number?

If that's not possible, perhaps at least dumping the problematic context?

PR #166, opened 2025-04-24
Human-friendly number parsing

sfc-gh-mwyatt:
Added `HumanInt` and `HumanFloat` types for the pydantic configs. Allows parsing human-friendly strings to `int` or `float` values. The following formats are supported:
- `1_000` --> `1000`
- `50%` --> `0.5`
- `1e-3` --> `0.001`
- `2^10` --> `1024`
- `4K` --> `4000` (also `M`, `B`, and `T`)
- `4Ki` --> `4096` (also `Mi`, `Gi`, and `Ti`)

### DIFF
--- a/arctic_training/config/checkpoint.py
+++ b/arctic_training/config/checkpoint.py
@@ -20,6 +20,7 @@ from typing import Type
 from pydantic import Field
 
 from arctic_training.config.base import BaseConfig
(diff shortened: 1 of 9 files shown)


sfc-gh-sbekman (approved):
Looks amazing, Mike! thank you!

Should we document somewhere which string formats are supported?

PR #169, opened 2025-04-25
Fix ReadTheDocs build and add unit test

sfc-gh-mwyatt:
Various changes to the code has caused warnings to surface in the docs build. Added fixes to address these and have properly building docs again. Also adding a unit test that builds the docs to avoid this problem moving forward.

### DIFF
--- a/.github/workflows/formatting.yaml
+++ b/.github/workflows/formatting.yaml
@@ -21,11 +21,6 @@ jobs:
         id: setup_python
         with:
           python-version: '3.10'
-      - name: Restore cached virtualenv
-        uses: actions/cache@v4
-        with:
-          key: venv-${{ runner.os }}-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('pyproject.toml') }}
-          path: .venv
       - name: Restore cached pre-commit environment
         uses: actions/cache@v4
         with:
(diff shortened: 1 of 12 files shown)


.github/workflows/formatting.yaml
```diff
@@ -21,23 +21,14 @@ jobs:
...
           path: ~/.cache/pre-commit
           key: pre-commit-${{ runner.os }}-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('.pre-commit-config.yaml') }}
       - name: Setup environment
         run: |
-          python -m venv .venv
-          source .venv/bin/activate
-          echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
-          echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> $GITHUB_ENV
-          python -m pip install ".[formatting]"
+          python scripts/install_deps.py formatting
```
sfc-gh-mwyatt, line 31: Turns out this is much faster than caching the environment!

arctic_training/config/enums.py
```diff
@@ -45,3 +46,19 @@ def __eq__(self, other) -> bool:
         if isinstance(other, DType):
             return self is other
         return other in self._aliases
+
+    @classmethod
+    def __get_pydantic_core_schema__(cls, source_type, _handler):
```
sfc-gh-mwyatt, line 51: Fixes for serialization issues with this custom type

arctic_training/config/model.py
```diff
@@ -52,29 +50,33 @@ class ModelConfig(BaseConfig):
...
-    peft_config: Optional[PeftConfig] = None
+    peft_config: Optional[Dict] = None
     """ Configuration for Parameter Efficient Fine Tuning. """
 
     @property
     def factory(self) -> Type["ModelFactory"]:
         return get_registered_model_factory(name=self.type)
 
+    @property
+    def peft_config_obj(self) -> peft.PeftConfig:
```
sfc-gh-mwyatt, line 61: Pydantic doesn't like arbitrary `dataclass` type objects, so just storing the dict in `peft_config` field and generating this object when it's needed.

scripts/install_deps.py
```diff
@@ -16,10 +16,14 @@
 import argparse
 import subprocess
 import sys
-import tomllib
 from typing import Dict
 from typing import List
 
+try:
+    import tomllib
```
sfc-gh-mwyatt, line 23: `python<3.11` does not have `tomllib` so adding this work-around

sfc-gh-jrasley (approved):

PR #174, opened 2025-05-05
Avoid corrupted data cache

### LINKED ISSUE #172: detecting an incomplete data cache
Currently if I kill AT while it's building a cache, it then often fails to read it on restart and I need to manually remove it. This could be easily solved by adding a flag file like `ready.txt` that gets written at the end of cache writing. If it's there we know the cache is valid, if it is not there we know we shouldn't attempt to load it and need to build one.

If for whatever reason we are re-building cache that already has `ready.txt` we need to delete it first.

sfc-gh-mwyatt:
Resolves #172 

Write to `{cache_path}.incomplete` and rename after saving all data is complete.


sfc-gh-sbekman (approved):
I like the way you solved this, Mike!

PR #175, opened 2025-05-05
Add early exit kill switch

### LINKED ISSUE #126: [feature request] Kill switch
Another very useful training feature is a kill-switch. It allows an early exit at will - the earliest point is the end of the current iteration.

This is helpful when starting a training, realizing you can't wait till its over but wanting a clean exit and not Ctrl-C, so that say a checkpoint gets saved or a memory profiler logs its results.

Currently I started a long sequence 10-step training, didn't realize it'll take 50min to finish, but it has already run for 20min so I don't want to waste another 20min restarting and want to have a clean exit, but forgot AT doesn't yet have a kill-switch.

1. config:
```
kill_switch_path : output_dir/kill-switch.txt
```
2. it's just another signal for an early exit in a training loop, pseudo code:

```
for batch in dl:
  step()
  if self.kill_switch_path.exists(): 
    self.training_finished = True
    self.kill_switch_path.remove()
  exit if self.training_finished
```

3. To arm the kill switch, the user runs:

```
touch /path/tooutput_dir/kill-switch.txt
```
when they want the trainer to exit once this running iteration finishes.

sfc-gh-mwyatt:
Allow user to manually set `early_stop=True` during training by creating a file. Defaults to `/tmp/at_kill_switch`, but can be modified with `kill_switch_path` in the trainer config.

Resolves #126

### DIFF
--- a/arctic_training/trainer/trainer.py
+++ b/arctic_training/trainer/trainer.py
@@ -333,6 +333,9 @@ class Trainer(ABC, CallbackMixin, metaclass=RegistryMeta):
                         step=self.model.global_steps,
                     )
 
+            if self.config.kill_switch_path.exists():
+                self.early_stop = True
+
             if self.early_stop:
                 break
         self.metrics.stop_timer("iter")
--- a/arctic_training/config/trainer.py
+++ b/arctic_training/config/trainer.py
@@ -143,6 +143,9 @@ class TrainerConfig(BaseConfig):
     mem_profiler_max_entries: HumanInt = Field(default=100_000, ge=1)
     """ Maximum number of entries to store in the memory profiler. """
 
+    kill_switch_path: Path = Path("/tmp/at_kill_switch")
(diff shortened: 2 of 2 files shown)


sfc-gh-jrasley (approved):

arctic_training/trainer/trainer.py
```diff
@@ -333,6 +333,9 @@ def epoch(self) -> None:
                         step=self.model.global_steps,
                     )
 
+            if self.config.kill_switch_path.exists():
+                self.early_stop = True
```
sfc-gh-sbekman, line 337: This isn't going to work well in multi-node situation, as you're checking one node's fs. So this will require a shared fs.

I think the above needs to check for global rank 0's fs if it's local and then signal to other nodes to exit - otherwise it'll break.

If you want a quick solution make the default local to where the code launched from and perhaps use our util that maps the fs type to ensure that it lands on a shared fs. If not, assert until signaling is implemented.

PR #178, opened 2025-05-08
Add data-process mode to CLI

sfc-gh-mwyatt:
Maintains current CLI of `arctic_training config.yaml` but enables a new mode to do all data processing and then early-exit from trainer initialization. This can be used to process and cache data ahead of training runs.

Behavior:
- `arctic_training config.yaml` --> runs training
- `arctic_training train config.yaml` --> runs training
- `arctic_training process-data config.yaml` --> runs data processing, caches data, early exits

@sfc-gh-sbekman

### DIFF
--- a/arctic_training/cli.py
+++ b/arctic_training/cli.py
@@ -39,6 +39,14 @@ def main():
             """
         ),
     )
+    parser.add_argument(
+        "mode",
+        type=str,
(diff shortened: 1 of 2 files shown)


sfc-gh-sbekman (approved):
Thank you, Mike

PR #186, opened 2025-05-28
Add evaluation method

sfc-gh-mwyatt:


arctic_training/metrics.py
```diff
@@ -186,7 +190,10 @@ def print_summary(self) -> None:
             f" {100*self.summary_dict['iter']//self.max_iter:>3}%"
         )
         if "loss" in self.summary_dict:
-            summary_str += f" | loss: {self.summary_dict['loss']:.4f}"
+            loss_label = "loss" + ("/eval" if self.trainer.config.eval_frequency > 0 else "")
```
sfc-gh-sbekman, line 193: don't you have that label already logged 3 lines below but with a different value, so `loss/eval` appears twice?

arctic_training/metrics.py
```diff
@@ -186,7 +190,10 @@ def print_summary(self) -> None:
...
+            summary_str += f" | {loss_label}: {self.summary_dict['loss']:.4f}"
+        if "loss/eval" in self.summary_dict:
+            summary_str += f" | loss/eval: {self.summary_dict['loss/eval']:.4f}"
```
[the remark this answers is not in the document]
sfc-gh-prenc: Agree, the evaluation metrics should be logged separately. In theory, setting `train_log_iter_interval` to zero disables logging training metrics, but with the current implementation, this also suppresses evaluation metrics, which is just wrong.

arctic_training/config/trainer.py
```diff
@@ -142,6 +142,7 @@ class TrainerConfig(BaseConfig):
     """ Maximum number of training iterations. """
 
     eval_frequency: int = Field(default=0, ge=0)
```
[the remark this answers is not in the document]
sfc-gh-prenc: I think these refer to slightly different concepts. `loss_log_interval` and `train_log_iter_interval` control logging for all training steps, including intermediate ones. `eval_frequency` only applies to GAS iterations, since it wouldn’t make sense to run evaluation more frequently than that.

arctic_training/metrics.py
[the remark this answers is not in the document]
sfc-gh-sbekman: In all the previous frameworks I was involved in there was a separate: `eval_log_iter_interval` which removes any ambiguity.

arctic_training/metrics.py
sfc-gh-prenc: outdated

arctic_training/config/trainer.py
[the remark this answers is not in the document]
sfc-gh-prenc: To clarify, I’m referring to the conditions below. `train_log_iter_interval` is checked against `train_batch_idx`, which includes all training steps, including intermediate ones between gradient accumulation steps. Whereas, `eval_frequency` is checked against global_steps, and evaluation only happens at the gas_boundary.

Here are the relevant conditions:

Training condition:
```python
            if (
                self.config.train_log_iter_interval != 0
                and self.train_batch_idx % self.config.train_log_iter_interval == 0
            ):
                self.metrics.print_summary()
```

Evaluation condition:
```python
            if (
                self.gas_boundary
                and self.config.eval_frequency != 0
                and self.model.global_steps % self.config.eval_frequency == 0
            ):
                self.evaluate()
```

arctic_training/callback/default.py
```diff
@@ -16,7 +16,10 @@
 
 def validate_callback(self):
     if self.model.is_gradient_accumulation_boundary():
-        if self.config.eval_frequency > 0 and (self.model.global_steps + 1) % self.config.eval_frequency == 0:
+        if (
+            self.config.eval_log_iter_interval > 0
+            and (self.model.global_steps + 1) % self.config.eval_log_iter_interval == 0
+        ):
             self.validate()
```
sfc-gh-mwyatt, line 23: We can actually remove this as it's not used anywhere and feedback has generally been to avoid using callbacks in the base classes.

sfc-gh-prenc: Okay, I will remove the whole file.

arctic_training/config/trainer.py
```diff
@@ -141,7 +141,8 @@ class TrainerConfig(BaseConfig):
     train_iters: HumanInt = Field(default=0, ge=0)
     """ Maximum number of training iterations. """
 
-    eval_frequency: int = Field(default=0, ge=0)
+    eval_log_iter_interval: int = Field(default=0, ge=0)
```
sfc-gh-mwyatt, line 144: @sfc-gh-prenc I think we should still have `eval_frequency` and `eval_log_iter_interval`. `eval_log_iter_interval` should only control when eval results are logged (as indicated by the name). But `eval_frequency` (or `eval_iters` we can name it whatever) should control how often eval happens.

Additionally, we can make `eval_log_iter_interval` accept only `0` or `1` for the time being. Once we overhaul the metrics system to better aggregate and report metrics over larger intervals, we can allow larger values.

sfc-gh-sbekman: No objections to multiple configs, but I still maintain that the terms `frequency` and `interval` indicate the exact same thing, so let's choose one and stick to it. Otherwise it's impossible to remember when to use which.

sfc-gh-mwyatt: Agreed, `eval_interval` is probably best!

sfc-gh-sbekman:
> Every time, when I run evaluation, it triggers bunch of initializations which adds a significant overhead to the training time

Normally there should be only one trainer and it should be able to switch between the 2 modes w/o doing any re-init. So it's probably a bug or a design issue if the trainer is rebuilt every time.

The only change that should happen during the switch between the 2 modes is changing the model state from train to eval and the no_grad context.

The problem with having both eval and train usually comes from DL as now you have 2 sets, which could impose additional stress on memory usage, but I think in our case our DL is very light.

Finally, one other thing that may or may not be important is RNG which will be impacted by eval and will prevent reproducibility unless restored between each mode, but perhaps you already sorted this out.

sfc-gh-mwyatt:
> > Every time, when I run evaluation, it triggers bunch of initializations which adds a significant overhead to the training time
> 
> Normally there should be only one trainer and it should be able to switch between the 2 modes w/o doing any re-init. So it's probably a bug or a design issue if the trainer is rebuilt every time.
> 
> The only change that should happen during the switch between the 2 modes is changing the model state from train to eval and the no_grad context.
> 
> The problem with having both eval and train usually comes from DL as now you have 2 sets, which could impose additional stress on memory usage, but I think in our case our DL is very light.
> 
> Finally, one other thing that may or may not be important is RNG which will be impacted by eval and will prevent reproducibility unless restored between each mode, but perhaps you already sorted this out.

We sorted this out by adding `persistent_workers=True` to the dataloader kwargs. The dataloaders were spinning up new procs at the beginning of each iteration through the dataset!

With regards to RNG, this is a very good point and we should take a closer look.

arctic_training/config/trainer.py
```diff
@@ -316,9 +320,13 @@ def init_tokenizer_config(cls, v: Union[Dict, TokenizerConfig], info: Validation
         return cast(TokenizerConfig, subconfig)
 
     @model_validator(mode="after")
-    def validate_eval_frequency(self) -> Self:
+    def validate_eval_log_iter_interval(self) -> Self:
```
sfc-gh-mwyatt:
```suggestion
    def validate_eval_interval(self) -> Self:
```

arctic_training/config/trainer.py
```diff
@@ -316,9 +320,13 @@ def init_tokenizer_config(cls, v: Union[Dict, TokenizerConfig], info: Validation
...
         if self.data.eval_sources or self.data.train_eval_split[1] > 0.0:
-            assert self.eval_frequency > 0, "eval_frequency must be set if eval dataset is provided."
+            assert self.eval_interval > 0, "`eval_frequency` must be set if eval dataset is provided."
```
sfc-gh-sbekman, line 325:
```suggestion
            assert self.eval_interval > 0, "`eval_interval` must be set if eval dataset is provided."
```

sfc-gh-sbekman, line 325: why listing the eval dataset requires using it?

I think it's normal to have it provided by to use `eval_interval` setting to decide whether to eval or not.

arctic_training/config/trainer.py
```diff
@@ -316,9 +320,13 @@ def init_tokenizer_config(cls, v: Union[Dict, TokenizerConfig], info: Validation
...
+        if self.eval_interval > 0:
+            assert (
+                self.data.eval_sources or self.data.train_eval_split[1] > 0.0
+            ), "`eval_frequency` must be set only if eval dataset is provided."
```
sfc-gh-sbekman, line 329:
```suggestion
            ), "`eval_interval` must be set only if eval dataset is provided."
```

projects/arctic_embed/examples/finetune_models/finetune_e5_base_unsupervised.py
```diff
@@ -135,7 +135,7 @@ def configure_non_distributed_distributed_training_if_needed() -> None:
         wandb=wconf,
         deepspeed=dsconf,
         loss_log_interval=0,
-        eval_frequency=100,
+        eval_log_iter_interval=100,
```
sfc-gh-mwyatt, line 138:
```suggestion
        eval_interval=100,
        eval_log_iter_interval=100,
```

arctic_training/trainer/trainer.py
```diff
@@ -415,21 +419,32 @@ def epoch(self) -> None:
...
-                        metrics.pop("iter")  # not needed for wandb
-                        self.wandb_experiment.log(metrics, step=self.model.global_steps)
+                    if self.wandb_experiment is not None:
+                        metrics = {k: v for k, v in metrics.items() if k not in ["iter"]}
+                        self.wandb_experiment.log(metrics, step=self.global_step)
+
+                if self.config.eval_interval != 0 and self.global_step % self.config.eval_interval == 0:
+                    self.evaluate()
+
+                    if not self.eval_batches.disable:
```
sfc-gh-mwyatt, line 442: This could maybe impact performance since we have to create a new tqdm object over eval dataloader when we call `self.eval_batches` - but it's probably fine for now

arctic_training/trainer/trainer.py
```diff
@@ -415,21 +419,32 @@ def epoch(self) -> None:
 
             self.metrics.restart_timer("iter")
 
-            if (
-                self.config.train_log_iter_interval != 0
-                and self.train_batch_idx % self.config.train_log_iter_interval == 0
-            ):
+            if self.config.train_log_iter_interval != 0:
                 self.metrics.print_summary()
```
sfc-gh-sbekman, line 423: should we split print_summary into accumulate and print? otherwise it's confusing - it shouldn't print anything unless it's at the gas boundary, which is the next if branch below.

on the other hand I know you're planning a revamp of this so it's totally fine to leave it as it is for now.

sfc-gh-mwyatt (commented):
Just a few changes

arctic_training/trainer/trainer.py
```diff
@@ -465,6 +480,17 @@ def train(self) -> None:
...
                 self.wandb_experiment.finish()
 
+    @callback_wrapper("evaluate")
+    @torch.inference_mode()
+    def evaluate(self) -> None:
+        """
+        Evaluation loop. Measures the model's performance on the evaluation dataset.
+        """
+        self.model.eval()
+        losses = [self.loss(eval_batch).item() for eval_batch in self.eval_batches]
```
sfc-gh-sbekman, line 490: no need for `with torch.no_grad`?

projects/arctic_embed/examples/finetune_models/finetune_e5_base_unsupervised.py
sfc-gh-prenc: One thing: since `eval_log_iter_interval` is calculated based on evaluation iterations, if we evaluate every 100th iteration and want to log every one of them, we set `eval_log_iter_interval=1`. But if we want to log every second evaluation (i.e., every 200th global iteration), we set `eval_log_iter_interval=2`. This is how it’s supposed to work right now. Let me know what you think.

arctic_training/config/trainer.py
sfc-gh-prenc: This prevents the user from adding an eval dataset, starting the training, and only later realizing they forgot to turn on evaluation, potentially wasting both their time and the GPU’s time. The other side is what you mentioned: the annoying code commenting out. IMO, it’s not easy to decide which option we should go with. I guess we should choose the one that best balances the probability of happening and the potential time saved.

arctic_training/trainer/trainer.py
sfc-gh-prenc: We already sped it up a lot by fixing the worker reinitialization. This is minor, I’d leave it for the future, as you’re saying.

arctic_training/trainer/trainer.py
sfc-gh-prenc: In the initial implementation, it was printing every iteration (including intermediate ones), so I left it as is. Yeah, making it print only on each gas iteration would require more work to accumulate everything, which doesn’t make sense to spend time on now if the entire metrics system is planned to be overhauled. I’d leave it like this for now.

sfc-gh-sbekman: the training metrics already accumulate some but not all, hence the boundary flag.

But I'm totally fine to do it in stages, Pawel

sfc-gh-sbekman (approved):

arctic_training/config/trainer.py
sfc-gh-prenc: > the simple eval is often quite useless, the training loss is good enough of a signal, unless you do many epoch and try to avoid memorization - typically eval is done asynchroniously to training and usually involves quite a complex and rich set of evals and not just loss.

I generally agree. However, I don’t work in NLP, and in my domain of EHR (medical data), the data is scarce, which easily leads to overfitting. In my experience, the evaluation loss is the best indicator of performance, I barely even look at the training loss.

sfc-gh-sbekman: Totally understandable, Pawel.

In modern NLP training (especially pre-training) there is only ever 1 epoch so indeed your use case is totally different. I can imagine post-training might want to do multiple epochs as well.

Let's support both approaches.

PR #192, opened 2025-06-01
Add DeepseekV2SwiftKV and reorganize swiftkv project

sfc-gh-aqiao:


projects/swiftkv/configs/llama-3.1-swiftkv-405b-instruct.yaml
```diff
@@ -17,13 +17,11 @@ data:
...
     - meta-math/MetaMathQA
     - ise-uiuc/Magicoder-OSS-Instruct-75K
-  use_data_cache: true
-  cache_processed_data: true
   cache_dir: /data/data-cache
   num_proc: 16
   max_length: 5120
 logger:
-  level: WARNING
+  level: INFO
```
sfc-gh-mwyatt, line 24: Any reason to have such verbose logging?

sfc-gh-mwyatt (approved):

projects/swiftkv/configs/llama-3.1-swiftkv-405b-instruct.yaml
sfc-gh-aqiao: For SwiftKV it outputs several losses like student loss, teacher loss, which are useful but aren't in the less verbose logging

sfc-gh-mwyatt: I see - ok.

We have plans to update the metrics, such that custom trainers can add arbitrary metrics that get logged at each training iter. I will make a note to update so the student/teacher loss are reported there to avoid overly-verbose output. But this works for the time being!

PR #196, opened 2025-06-04
Long-context SwiftKV training

sfc-gh-aqiao:


sfc-gh-sbekman (approved):
LGTM, with some discussion on slack.

arctic_training/data/sft_factory.py
```diff
@@ -234,6 +245,17 @@ class SFTDataConfig(DataConfig):
     pack_samples: bool = True
     """ Whether to pack multiple samples into samples up to size `max_length`. """
 
+    drop_last: bool = False
+    """ Whether to drop the last packed sample, which might be shorter than `max_length`. """
+
+    fuse_positions_prob: float = 0.0
```
sfc-gh-mwyatt, line 251:
```suggestion
    fuse_positions_prob: float = Field(0.0, ge=0.0, le=1.0)
```

projects/swiftkv/train.py
```diff
@@ -216,6 +222,11 @@ def loss_fn(student_concat, teacher_logits, teacher_hidden, mask):
                 compute_params=[],
                 output_unshard_dimension=0,  # loss is a scalar
             )
+
+            # Average the loss across all sequence parallel ranks
```
sfc-gh-sbekman, line 226: I'd just expand the comment to say that this is a differentiable mean, so that in the future someone won't replace it with a normal all_reduce.

projects/swiftkv/configs/llama-3.1-swiftkv-8b-instruct-long.yaml
```diff
@@ -0,0 +1,88 @@
...
+  name_or_path: meta-llama/Llama-3.1-8B-Instruct
+  num_key_value_layers: 16
+  key_value_group_size: 1
+  attn_implementation: flash_attention_2
+deepspeed:
+  zero_optimization:
+    stage: 2
+data:
+  sources:
+    - type: Open-Orca/OpenOrca
```
sfc-gh-mwyatt, line 19: This is an unfortunate pattern for the `type` and `name_or_path`. We should probably auto-fill the `name_or_path` if the type refers to a specific sub-class of `DataSource`. I will create a follow-up PR that allows us to reduce the verbosity.

sfc-gh-mwyatt (approved):

projects/swiftkv/configs/llama-3.1-swiftkv-8b-instruct-long.yaml
sfc-gh-aqiao: Yep this part is a little awkward right now

PR #197, opened 2025-06-04
Add random sampling from data sources

sfc-gh-aqiao:


arctic_training/data/source.py
```diff
@@ -66,6 +67,23 @@ def __call__(self) -> DatasetType:
             return load_from_disk(self.cache_path.as_posix())
 
         dataset = self.load(self.config, self.config.split)
+
+        sample_count = None
+
+        if self.config.sample_ratio is not None:
+            assert self.config.sample_count is None, (
+                "sample_ratio and sample_count cannot both be set.")
```
sfc-gh-aqiao, line 75: @sfc-gh-mwyatt is this the right place to do this check?

arctic_training/data/source.py
```diff
@@ -66,6 +67,22 @@ def __call__(self) -> DatasetType:
             return load_from_disk(self.cache_path.as_posix())
 
         dataset = self.load(self.config, self.config.split)
+
+        sample_count = None
+
+        if self.config.sample_ratio is not None:
+            assert self.config.sample_count is None, "sample_ratio and sample_count cannot both be set."
```
sfc-gh-mwyatt, line 74: Can we move this to a pydantic validator? We should try to error out on any bad configs as early as possible to avoid wasting time loading other assets only to error out later!

```python
def sample_ratio_or_sample_count(self) -> Self:
    assert not (self.sample_ratio is not None and self.sample_count is not None), "sample_ratio and sample_count cannot both be set."
    # Or alternatively:
    if self.sample_ratio is not None and self.sample_count is not None:
        assert False, "sample_ratio and sample_count cannot both be set"
```

arctic_training/data/source.py
sfc-gh-mwyatt: See my comment below

sfc-gh-mwyatt (approved):

PR #199, opened 2025-06-05
fix lr scheduler for sequence parallel

sfc-gh-aqiao:
Currently using Nx sequence parallel the lr scheduler will advance Nx too fast, hitting lr=0 after 1/N of the training. This PR fixes that.


sfc-gh-aqiao:
@sfc-gh-sbekman @sfc-gh-mwyatt

sfc-gh-mwyatt (approved):

PR #204, opened 2025-06-12
Fix custom code import error

sfc-gh-mwyatt:
When using multiple workers on the torch `DataLoader` object, if a user has custom data classes defined in their projects `train.py`, we would get an import error like `ModuleNotFoundError: No module named 'custom_trainer'` from the spawned child processes.

Fixing that by making sure that the custom user objects are importable to child processes. This is achieved by creating a temporary directory in `/tmp/` and then creating a symlink to the users custom objects script (e.g., `/tmp/user_trainer_symlinks/user_train_828a1303.py` --> `./train.py`). This unique name is to avoid any potential conflicts with other modules.

### DIFF
--- a/arctic_training/config/trainer.py
+++ b/arctic_training/config/trainer.py
@@ -15,6 +15,8 @@
 
 import importlib.util
 import sys
+import tempfile
+import uuid
 from functools import partial
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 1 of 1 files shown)


sfc-gh-jrasley (approved):

sfc-gh-jrasley:
definitely adds more complexity to the magic custom trainer path but glad it fixes the core issue.

PR #222, opened 2025-06-26
Fix for custom user script relative import

sfc-gh-mwyatt:
If a user has a `train.py` with a relative import:
```python
from my_local_util import my_custom_util_func
...
```

Then this import would fail as only the `train.py` is symlinked to the custom module directory. Instead creating a symlink for the entire directory where the custom code is located. This fixed the problem and allows relative imports.

Added a test to verify this functionality.

@sfc-gh-prenc this is a follow up to the problem we fixed in #204

### DIFF
--- a/arctic_training/config/trainer.py
+++ b/arctic_training/config/trainer.py
@@ -406,26 +406,35 @@ class TrainerConfig(BaseConfig):
 
 
 def load_user_module_from_path(script_path: Path) -> None:
-    # Symlink the script to a temporary directory to avoid clashing with other modules
+    # Symlink the entire directory containing the script to avoid issues with relative imports
+    script_dir = script_path.parent
     tmp_root = Path(tempfile.gettempdir())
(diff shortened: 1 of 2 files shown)


sfc-gh-prenc:
@sfc-gh-mwyatt Training crashes:
```bash
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=121, pipe_handle=141)
                                                  ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/prenc/miniconda/lib/python3.13/multiprocessing/spawn.py", line 122, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "/data/prenc/miniconda/lib/python3.13/multiprocessing/spawn.py", line 132, in _main
    self = reduction.pickle.load(from_parent)
ModuleNotFoundError: No module named 'user_dir_9e9ef83a_train'
```

PR #229, opened 2025-07-02
Switch swiftkv llama-70b base model from 3.1 to 3.3

sfc-gh-jrasley:
Align 70b recipe with what we just ran in prod (decreased max length and also used llama-3.3)


projects/swiftkv/configs/llama-3.3-swiftkv-70b-instruct-long.yaml
```diff
@@ -7,7 +7,7 @@ sequence_parallel_size: 32
 logits_loss_temp: 2.0
 hidden_loss_layer: -2
 model:
-  name_or_path: meta-llama/Llama-3.1-70B-Instruct
+  name_or_path: meta-llama/Llama-3.3-70B-Instruct
```
sfc-gh-aqiao, line 10: rename the file too or add this as a different config file?

PR #239, opened 2025-07-15
report the correct device in pynvml

sfc-gh-sbekman:
Currently if I run `CUDA_VISIBLE_DEVICES=2 arctic_training ...` the memory usage in metrics logs is reported for gpu0 and not gpu2, so adding some logic to figure the device id from the env vars.


sfc-gh-mwyatt (approved):

sfc-gh-sbekman:
Thank you, Mike.

PR #252, opened 2025-08-01
checkpoint resume support

sfc-gh-jrasley:
Tested with llama-1b on 1 and 2 gpus, with and without swiftkv with the config below. Ran training for ~20 steps saving losses, model weight norm, optim state norms. Resume from step 10 and confirmed steps 11-20 are the same as the original run. Needed to test with an `eager` attn implementation in order to get deterministic loss values.

```yaml
type: swiftkv
code: ../train.py
micro_batch_size: 1
epochs: 1
gradient_accumulation_steps: 1
model:
  name_or_path: meta-llama/Llama-3.2-1B-Instruct
  num_key_value_layers: 8
  key_value_group_size: 1
  attn_implementation: eager
deepspeed:
  zero_optimization:
    stage: 2
data:
  sources:
    - type: HuggingFaceH4/ultrachat_200k
      name_or_path: HuggingFaceH4/ultrachat_200k
      sample_count: 200
  cache_dir: /data-fast/data-cache
  num_proc: 16
  max_length: 8192
logger:
  level: INFO
  output_dir: "./"
  file_output_ranks: [0]
optimizer:
  betas: [0.9,0.999]
  weight_decay: 0.0
  lr: 0.1
checkpoint:
  - type: deepspeed
    save_every_n_epochs: 1
    save_every_n_steps: 10
    output_dir: /data-fast/ckpt-swiftkv-test
    save_end_of_training: false
    auto_resume: true
```


arctic_training/checkpoint/ds_engine.py
```diff
@@ -24,9 +24,39 @@
 from arctic_training.checkpoint.engine import CheckpointEngine
 
 
+def model_norm(model: torch.nn.Module) -> float:
+    with torch.no_grad():
+        all_params = torch.cat([p.view(-1) for p in model.parameters()])
```
sfc-gh-sbekman, line 29: won't this push into OOM if the memory usage was snag in the first place, as this may end up being a huge copy, no? perhaps running norm on each and then averaging?

same goes for optm states concat next.

I do see that it's a debug code...

arctic_training/trainer/trainer.py
```diff
@@ -375,7 +379,7 @@ def maybe_item(v):
 
         self.model.step()
 
-        # use deepspeed global step as golden truth
+        # DeepSpeed incremented its global step after the step() call, so we use it as the golden truth
```
sfc-gh-sbekman, line 382:
```suggestion
        # DeepSpeed increments its global step after the step() call, so we use it as the golden truth
```

arctic_training/trainer/trainer.py
```diff
@@ -400,7 +404,22 @@ def epoch(self) -> None:
...
 
-        for batch in self.train_batches:
+        batch_iterator = iter(self.train_batches)
+        if self._resumed_from_checkpoint:
+            logger.info(f"Resumed from checkpoint at global step: {self.global_step}.")
+            batches_to_skip = self.global_step % len(self.train_dataloader)
+            logger.info(f"Advancing {batches_to_skip} batches.")
+            for _ in range(batches_to_skip):
+                next(batch_iterator)
+                self.train_batch_idx += 1
```
sfc-gh-sbekman, line 414:
```suggestion
            self.train_batch_idx += batches_to_skip
```

arctic_training/trainer/trainer.py
```diff
@@ -400,7 +404,22 @@ def epoch(self) -> None:
...
+            self._resumed_from_checkpoint = False
+
+        for batch in batch_iterator:
+            if self._resumed_from_checkpoint:
+                batches_to_skip -= 1
+                if batches_to_skip <= 0:
+                    self._resumed_from_checkpoint = False
+                    logger.info("Resumed from checkpoint complete.")
```
sfc-gh-sbekman, line 422: could you please explain what is the purpose of this code? you already skipped all the batches earlier, no?

sfc-gh-sbekman (approved):
Looks good, Jeff. Added a few possible suggestions, but feel free to ignore.

arctic_training/trainer/trainer.py
sfc-gh-jrasley: Oh good catch, thank you! I left in this 2nd block on accident. I'll remove it as you are right it's never triggered.

sfc-gh-sbekman:
Looks good now, Jeff.

PR #260, opened 2025-08-12
Speculator for gpt-oss

sfc-gh-jaelee:


sfc-gh-jrasley (approved):

sfc-gh-jrasley:
@sfc-gh-jaelee @sfc-gh-aqiao @sfc-gh-yewang we really need to update the readme under `projects/arctic_lstm_speculator` to explain how to use all of this, we don't really have any documentation here other than blogs which are frozen in time after they are released.

PR #269, opened 2025-08-22
liger-kernel: cleanly report when a model isn't supported

sfc-gh-sbekman:
with this PR when liger-kernel doesn't support a model arch, it'll throw a user-friendly exception, like:
```
[rank6]: ValueError: It appears that liger-kernel==0.6.2 doesn't support the architecture of 
ibm-granite/granite-3.0-3b-a800m-instruct, perhaps try the latest liger-kernel version. 
It can't find 'granitemoe'.
```

### DIFF
--- a/arctic_training_cli.py
+++ b/arctic_training_cli.py
@@ -15,6 +15,7 @@
 
 
 import argparse
+import os
 import shutil
 import textwrap
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


sfc-gh-sbekman:
@sfc-gh-mwyatt ? let's merge this one? should take a sec to review

Issue #270, opened 2025-08-26
replace hardcoded zero params with values that depend on the model's size


sfc-gh-sbekman:
For optimal performance replace hardcoded zero params with values that depend on the model's size.

The current defaults are only well suited for a single model size:

```
ds_config["zero_optimization"] = ds_config.get(
            "zero_optimization",
            {
                "stage": 2,
                "stage3_param_persistence_threshold": 1e4,
                "stage3_max_live_parameters": 3e7,
                "stage3_prefetch_bucket_size": 3e7,
                "memory_efficient_linear": False,
            },
        )
```

Samyam's recommendations were (in HF Accelerate code):
```
                config_kwargs.update(
                    {
                        "zero_optimization.reduce_bucket_size": hidden_size * hidden_size,
                        "zero_optimization.stage3_prefetch_bucket_size": int(0.9 * hidden_size * hidden_size),
                        "zero_optimization.stage3_param_persistence_threshold": 10 * hidden_size,
                    }
```

PR #271, opened 2025-08-29
change deepspeed defaults

sfc-gh-sbekman:
1. the defaults are now model size-specific
2. removing `memory_efficient_linear` default to have user control it - no reason to disable it by default
3. adding `reduce_bucket_size` default which is important to set depending on the model size

Fixes: https://github.com/snowflakedb/ArcticTraining/issues/270

### DIFF
--- a/arctic_training/model/liger_factory.py
+++ b/arctic_training/model/liger_factory.py
@@ -42,10 +42,18 @@ class LigerModelFactory(HFModelFactory):
         swiglu = False if self.trainer.config.tiled_mlp_compute else True
         # XXX: it might be possible to combine the 2 in the future to benefit from the efficient liger swiglu kernel, but currently liger monkey patches the MLP class and thus we would have a race condition on who gets the override.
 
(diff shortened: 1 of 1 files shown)


arctic_training/config/trainer.py
```diff
@@ -386,14 +386,30 @@ def build_deepspeed_config(self) -> Self:
         ds_config["gradient_accumulation_steps"] = self.gradient_accumulation_steps
         ds_config["sequence_parallel_size"] = self.sequence_parallel_size
         ds_config["steps_per_print"] = ds_config.get("steps_per_print", 10)
+
+        from transformers import AutoConfig
+
+        model_config = AutoConfig.from_pretrained(self.model.name_or_path)
```
sfc-gh-sbekman, line 392: this is already done a few times in this file, perhaps it should be pre-run once and re-used?

I'm not sure if we have a hardcoded dependency on transformers - if we do, then it could be done.

PR #344, opened 2026-01-28
new feature: On policy distillation

sfc-gh-thonguyen:
Based on this blog post https://thinkingmachines.ai/blog/on-policy-distillation/ -- figured Arctic Training would be an appropriate place to have this feature. 

Training validated with GSM8K dataset on Qwen3-1.7B model using Qwen3-8B teacher.
image
image
Lower teacher perplexity means teacher is less surprised by the student's answer. Higher teacher logprob means teacher agrees with the student's answer.

image

Lower reverse KL means student's answers converge to teacher's.

image

Student's perplexity initially jumped up, meaning it's learning, and slowly ramped down, meaning it's getting more confident along the training progress.

Full dashboard: https://snowflake.wandb.io/thongnguyen/on-policy-distillation-gsm8k/runs/zuwzrd11?nw=nwuserthongnguyen

Once this PR is in we can make the claim ArcticTraining supports RL :)

### DIFF
--- a/arctic_training/debug.py
+++ b/arctic_training/debug.py
@@ -71,8 +71,12 @@ def get_nvml_mem():
             return 0
         pynvml_handle = pynvml.nvmlDeviceGetHandleByIndex(device_id)
         # pynvml.nvmlShutdown()
-    memory_info = pynvml.nvmlDeviceGetMemoryInfo(pynvml_handle)
-    return memory_info.used
+    try:
+        memory_info = pynvml.nvmlDeviceGetMemoryInfo(pynvml_handle)
+        return memory_info.used
+    except pynvml.NVMLError_NotSupported:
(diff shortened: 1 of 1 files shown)


arctic_training/trainer/on_policy_distillation_trainer.py
```diff
@@ -0,0 +1,520 @@
...
+
+        self.metrics.record("loss", maybe_item(loss))
+
+        self.model.step()
+
+        self.checkpoint()
+
+        # Update step counters
+        self.global_step = self.model.global_steps
+        self.global_step_this_run = self.global_step - self.global_step_at_start_this_run
```
sfc-gh-mwyatt: This appears to be an exact copy of the `step` method in the base trainer class. Why do we redefine it here?

